PR #19, opened 2020-04-20
API docs for grpc client + remove session duplication

chbatey:
I looked into each of the method to see if there is any statuss we
should handle but it was only execute batch dml.

I left streaming sql as not to conflict, do you want to do with @johanandren ?


chbatey:
As in i didn't look into if there was any status codes in the streaming query we should be dealing with

Issue #11, opened 2020-04-17
Handle chunked values in streaming queries


chbatey:
Particularly likely for the payload field

PR #21, opened 2020-04-20
Recombine chunked values

johanandren:
References #11


journal/src/main/scala/akka/persistence/spanner/internal/PartialResultSetFlow.scala
```diff
@@ -0,0 +1,93 @@
...
+import akka.util.OptionVal
+import com.google.protobuf.struct.Value.Kind
+import com.google.protobuf.struct.{ListValue, Value}
+import com.google.spanner.v1.PartialResultSet
+
+/**
+ * INTERNAL API
+ */
+@InternalApi object PartialResultSetFlow extends GraphStage[FlowShape[PartialResultSet, PartialResultSet]] {
+  // Custom graphstage avoids alloc per element in the happy case required by using statefulMapConcat
```
chbatey, line 16: 👍 every stateful map concat eventually becomes a stage!

journal/src/main/scala/akka/persistence/spanner/internal/PartialResultSetFlow.scala
```diff
@@ -0,0 +1,93 @@
...
+            previousChunk = OptionVal.None
+            val combined = recombine(previous, current)
+            if (combined.chunkedValue) {
+              previousChunk = OptionVal.Some(combined)
+              pull(in)
+            } else {
+              push(out, combined)
+            }
+          case OptionVal.None =>
+            if (current.chunkedValue) {
```
chbatey, line 40: Many rows can be in a single PartialResultSet, so i think we could have a full row here ready to push but the last value is chunked.

journal/src/main/scala/akka/persistence/spanner/internal/PartialResultSetFlow.scala
```diff
@@ -0,0 +1,93 @@
...
+@InternalApi object PartialResultSetFlow extends GraphStage[FlowShape[PartialResultSet, PartialResultSet]] {
+  // Custom graphstage avoids alloc per element in the happy case required by using statefulMapConcat
+  val in = Inlet[PartialResultSet](Logging.simpleName(this) + ".in")
+  val out = Outlet[PartialResultSet](Logging.simpleName(this) + ".out")
+
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with InHandler with OutHandler with StageLogging {
+      private var previousChunk: OptionVal[PartialResultSet] = OptionVal.None
```
chbatey, line 24: i think we should deliver what we can out of a result set and then keep just the values left over for the next row which may also have a chunk at the end

johanandren:
Hmm, ok, I was hoping we can keep the column/row collection and field dechunk as separate logics but I guess we need to combine those two then.

journal/src/main/scala/akka/persistence/spanner/internal/RowCollector.scala
```diff
@@ -0,0 +1,43 @@
...
+      {
+        var previousPartialRow: Seq[Value] = Nil
+        var columns: Seq[StructType.Field] = null
+
+        partialResultSet => {
+          // it is the first row, will contain metadata for columns
+          if (columns == null) {
+            columns = partialResultSet.metadata.get.rowType.get.fields
+          }
+          assert(columns != null, "received a row without first receiving metadata")
```
patriknw, line 26: (at least for historical reasons) we have been careful to not use `assert`, and rather IllegalStateException

johanandren:
I've combined field "unchunking" and extracting complete rows and emitting those as early as possible (even if there is a chunked field in the end of a resultset), into a single stage now.

Probably need more test cases though.

journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
```diff
@@ -246,7 +247,7 @@ private[spanner] class SpannerInteractions(spannerGrpcClient: SpannerGrpcClient,
         ),
         paramTypes = Schema.ReplayTypes
       )
-      .runForeach(row => replayRow(row))
-      .map(_ => ())
+      .runForeach(replayRow)
+      .map(ConstantFun.scalaAnyToUnit)
```
chbatey, line 251: 👍

journal/src/main/scala/akka/persistence/spanner/internal/RowCollector.scala
```diff
@@ -0,0 +1,153 @@
...
+import com.google.protobuf.struct.{ListValue, Value}
+import com.google.spanner.v1.PartialResultSet
+
+/**
+ * INTERNAL API
+ *
+ * Recombines chunked and incomplete `PartialResultSet`s coming from upstream into rows
+ * (represented by a `Seq` each) passed downstream
+ */
+@InternalApi private[akka] object RowCollector extends GraphStage[FlowShape[PartialResultSet, Seq[Value]]] {
```
chbatey, line 22: Looks good

johanandren:
I think it needs a few more unit tests and then running it against the simulator with some random table contents to see there isn't some scenario I didn't think of.

johanandren:
Added a bit more complicated resultset scenario mixing partial rows and chunked fields. I think that will do for now.

chbatey:
Nice!

Issue #12, opened 2020-04-17
SessionPool: Lifecycle of connections


chbatey:
Session pool needs to:
- [x] delete sessions on shut down
- [x] keep sessions alive (google defaults to closing them with an hour of inactivity)
- [x] repair broken sessions
- [x] consider timing out on client side or in the pool

chbatey:
The last thing that needs doing is to repair broken sessions. I'll do that now.

PR #22, opened 2020-04-20
Session pool supervision and cleanup

chbatey:
Part of #12

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
@@ -6,11 +6,11 @@ package akka.persistence.spanner.internal
 import java.util.UUID
 
 import akka.actor.typed.scaladsl.{AbstractBehavior, ActorContext, Behaviors, StashBuffer, TimerScheduler}
-import akka.actor.typed.{ActorRef, Behavior}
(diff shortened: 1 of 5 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -148,4 +144,18 @@ private[spanner] final class SessionPool(
...
+      cleanupOldSessions()
+      Behaviors.same
+    case PreRestart =>
+      cleanupOldSessions()
+      Behaviors.same
+  }
+
+  private def cleanupOldSessions(): Unit =
+    (availableSessions ++ inUseSessions.values).foreach { session =>
+      client.deleteSession(DeleteSessionRequest(session.name))
```
patriknw, line 159: How does this work for `inUseSessions`? I'm thinking about if it can be any problems when all actors are stopped during ActorSystem shutdown and there could be something in flight?

chbatey: I'll have to do some testing, I'd imagine all the queries will be aborted. A more complicated alternative would be to wait for all the sessions to be returned

chbatey: Created https://github.com/akka/akka-persistence-spanner/issues/23 to follow up. I'll do some testing against a real spanner at the end of the week before we cut a release

PR #24, opened 2020-04-20
Session pool keep alive

chbatey:
Refs #12


journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -102,20 +106,22 @@ private[spanner] final class SessionPool(
...
     ctx: ActorContext[Command],
     timers: TimerScheduler[Command],
-    stash: StashBuffer[Command]
+    stash: StashBuffer[Command],
+    settings: SessionPoolSettings
 ) extends AbstractBehavior[SessionPool.Command](ctx) {
+  private val keepAliveInMillis = settings.keepAliveInterval.toMillis
   private val log = ctx.log
-  private var availableSessions = initialSessions
+  private var availableSessions: List[(Session, Long)] = initialSessions.map((_, System.currentTimeMillis()))
```
johanandren, line 114: Should we perhaps keep them in a queue so we pick oldest for every new query, minimizing the need for `KeepAlive` at all for any given session?

Also, worth a case class with field names instead of a tuple?

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -140,6 +146,34 @@ private[spanner] final class SessionPool(
...
+        case (session, lastUse) if (currentTime - lastUse) > keepAliveInMillis =>
+          session
+      }
+      if (toKeepAlive.nonEmpty && log.isInfoEnabled) {
+        log.info(
+          "The following sessions haven't been used in the last {}. Sending keep alive. {}",
+          settings.keepAliveInterval.pretty,
+          toKeepAlive
+        )
+      }
```
johanandren, line 161: fold it into the if below?

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -140,6 +146,34 @@ private[spanner] final class SessionPool(
...
+          ctx.pipeToSelf(client.executeSql(ExecuteSqlRequest(session.name, sql = "SELECT 1"))) {
+            case Success(_) =>
+              ReleaseSession(id)
+            case Failure(t) =>
+              log.warn("failed to keep session alive, may be re-tried again before expires server side", t)
+              ReleaseSession(id)
+          }
+        }
+      }
+      Behaviors.same
```
johanandren, line 176: this

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
chbatey: Yes i'll do that

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
chbatey: I really do always forget to use this

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
chbatey: 👍

PR #46, opened 2020-04-22
Initial query journal

chbatey:
Doesn't implement anything yet, just so we can work on the different queries in parallel 

Also includes creating a single grpc client across all parts of the plugin

### DIFF
--- a/journal/src/main/resources/reference.conf
+++ b/journal/src/main/resources/reference.conf
@@ -1,9 +1,20 @@
 akka.persistence.spanner {
- class = "akka.persistence.spanner.SpannerJournal"
 
  project = "akka"
  instance = "akka"
  database = "akka"
+
+ journal {
+  class = "akka.persistence.spanner.SpannerJournal"
+ }
+
+ query {
+  class = "akka.persistence.spanner.SpannerReadJournalProvider"
+ }
+
+ snapshot {
+
+ }
 
  # name of the table to use for events
  table = "journal"
--- a/journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
+++ b/journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
@@ -7,7 +7,7 @@ package akka.persistence.spanner
 import akka.actor.ActorSystem
 import akka.event.Logging
 import akka.grpc.GrpcClientSettings
-import akka.testkit.TestKitBase
+import akka.testkit.{TestKit, TestKitBase}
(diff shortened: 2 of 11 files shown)


journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
```diff
@@ -259,12 +259,13 @@ trait SpannerLifecycle
...
-    cleanup()
-    // not sure why this doesn't get called by super.afterAll
-    TestKit.shutdownActorSystem(system)
-    super.afterAll()
-  }
+  override protected def afterAll(): Unit =
+    try {
+      cleanup()
+    } finally {
+      TestKit.shutdownActorSystem(system)
```
chbatey, line 266: because super.afterAll doesn't do it as this isn't typed so it doesn't know about scalatest

journal/src/main/resources/reference.conf
```diff
@@ -1,10 +1,21 @@
 akka.persistence.spanner {
- class = "akka.persistence.spanner.SpannerJournal"
 
  project = "akka"
  instance = "akka"
  database = "akka"
 
+ journal {
+  class = "akka.persistence.spanner.SpannerJournal"
```
chbatey, line 8: all the config is common right now

johanandren: Comparing with Couchbase we did a scope for the "connection" (project + instance + database in this case) I think the idea was to be able to change/override it separately for journal/query/snapshot by inheriting (`journal.connection = ${connection}`), but I don't see that it actually was implemented so maybe not useful enough.

chbatey: WDYT about just doing this for now? It's all internal and we can add more flexibility when needed

johanandren: Yep, sounds good not overdoing it.

journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
patriknw: why not use Typed `ScalaTestWithActorTestKit`?

chbatey: In the next PR i switch to using the ActorTestKit from typed but not ScalaTestWithActorTestKit as it is a class and we want most of the functionality in a trait so it can be used for the Journal specs

Issue #18, opened 2020-04-20
Retrying aborted write transactions


chbatey:
The commit of the write returns an gRPC ABORTED if the transaction fails. The advice is to re-try the transaction on the same session. For that reason I think we should re-try writes within the SpannerGrpcClient so the same session can be used

chbatey:
Maybe not for MVP?

PR #48, opened 2020-04-22
Retry writes on ABORTED failure

johanandren:
References #18

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -43,6 +43,7 @@ private[spanner] final class SpannerSettings(config: Config) {
   val table = config.getString("table")
   val deletionsTable = config.getString("deletions-table")
   val grpcClient = config.getString("grpc-client")
+  val writeRetries = 3 // FIXME from config
 
   val sessionPool = new SessionPoolSettings(config.getConfig("session-pool"))
   val sessionAcquisitionTimeout = config.getDuration("session-acquisition-timeout").asScala
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -46,7 +46,7 @@ private[spanner] final class SpannerSettings(config: Config) {
   val deletionsTable = config.getString("deletions-table")
   val grpcClient = config.getString("grpc-client")
   val maxWriteRetries = config.getInt("max-write-retries")
-  val maxWriteRetryInterval = config.getDuration("max-write-retry-interval").asScala
+  val maxWriteRetryTimeout = config.getDuration("max-write-retry-timeout").asScala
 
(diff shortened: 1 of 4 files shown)


johanandren:
I haven't figured out a way to verify this is exactly how the aborted exception will look but I think this is it.

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -43,6 +43,7 @@ private[spanner] final class SpannerSettings(config: Config) {
   val table = config.getString("table")
   val deletionsTable = config.getString("deletions-table")
   val grpcClient = config.getString("grpc-client")
+  val writeRetries = 3 // FIXME from config
```
chbatey, line 46: forget not

journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
```diff
@@ -83,22 +80,29 @@ private[spanner] object SpannerGrpcClient {
...
+              Transaction.SingleUseTransaction(
+                TransactionOptions(TransactionOptions.Mode.ReadWrite(TransactionOptions.ReadWrite()))
+              ),
+              mutations
+            )
+          )
+          .recoverWith {
+            case ex: StatusRuntimeException if ex.getStatus == io.grpc.Status.ABORTED && retriesLeft > 0 =>
+              log.debug("Write failed for [{}], retrying", session.id)
+              tryWrite(retriesLeft - 1)
```
chbatey, line 101: Looks good, simpler than the delete

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
johanandren: Yeah, wanted to avoid conflicting with your config changes until that is merged

johanandren:
I'll do the delete as well.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
```diff
@@ -83,22 +80,29 @@ private[spanner] object SpannerGrpcClient {
...
+            CommitRequest(
+              session.session.name,
+              Transaction.SingleUseTransaction(
+                TransactionOptions(TransactionOptions.Mode.ReadWrite(TransactionOptions.ReadWrite()))
+              ),
+              mutations
+            )
+          )
+          .recoverWith {
+            case ex: StatusRuntimeException if ex.getStatus == io.grpc.Status.ABORTED && retriesLeft > 0 =>
```
patriknw, line 99: btw, one thing to watch out for is if we get here due to long timeout. Then it would not be good to retry because circuit breaker might have triggered, actors restarted and so on.

johanandren: Ok, so perhaps we should also take a nanotime when it starts and take that into account when deciding to retry

Issue #45, opened 2020-04-22
Reuse the same session across a journal call


johanandren:
Right now we have hidden session acquisition in each query call in the `SpannerInteractions` meaning that each query will do an ask for a session, I think we should rather get a session up front in the journal methods and then re-use that session for the operations (which are anyway sequential for one call?)

chbatey:
👍

PR #49, opened 2020-04-23
Reuse session

johanandren:
Re-use session for multi query operations instead of a return-renew for each, use long for session request id.

References #45

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
@@ -34,18 +34,21 @@ import scala.util.{Failure, Success}
 @InternalApi
 private[spanner] object SessionPool {
   sealed trait Command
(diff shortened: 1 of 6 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -173,22 +184,31 @@ private[spanner] final class SessionPool(
...
+      stash.unstash(this, 1, identity)
+      this
+
+    case ThrowAwayAfterKeepAliveFail(session, cause) =>
+      log.warn(
+        s"Failed to keep session [${session.name}] alive, may be re-tried again before expires server side",
+        cause
+      )
+      // FIXME is this really the right thing to do?
+      availableSessions.enqueue(AvailableSession(session, System.currentTimeMillis()))
```
chbatey, line 208: Could have been a transient failure. Until we re-create them I think so

PR #53, opened 2020-04-24
Events by tag & all persistence ids

chbatey:
#7

### DIFF
--- a/journal/src/main/resources/reference.conf
+++ b/journal/src/main/resources/reference.conf
@@ -10,6 +10,10 @@ akka.persistence.spanner {
 
  query {
   class = "akka.persistence.spanner.SpannerReadJournalProvider"
+
+  # When live queries return no results. How often to poll Spanner for new
+  # rows
+  refresh-interval = 3s
  }
 
  snapshot {
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
@@ -4,6 +4,8 @@
 
 package akka.persistence.spanner
 
+import java.time.{Instant, ZoneOffset}
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
 import java.util.Base64
 
 import akka.actor.ActorSystem
@@ -37,6 +39,39 @@ private[spanner] object Schema {
     val Tags = "tags" -> Type(TypeCode.ARRAY)
 
     val Columns = List(PersistenceId, SeqNr, Event, SerId, SerManifest, WriteTime, WriterUUID, Tags).map(_._1)
+
+    val formatter = (new DateTimeFormatterBuilder)
+      .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+      .optionalStart
+      .appendOffsetId
+      .optionalEnd
+      .toFormatter
+      .withZone(ZoneOffset.UTC)
+
+    val NoOffset = formatter.format(Instant.ofEpochMilli(0))
+
+    def deserializeRow(serialization: Serialization, values: Seq[Value]): (PersistentRepr, String) = {
+      val persistenceId = values.head.getStringValue
+      val sequenceNr = values(1).getStringValue.toLong
+      val payloadAsString = values(2).getStringValue
+      val serId = values(3).getStringValue.toInt
+      val serManifest = values(4).getStringValue
+      // keep this in the offset as the original format rather than do any conversions
+      val writeOriginal = values(5).getStringValue
+      val writeTimestamp: Instant = Instant.from(formatter.parse(writeOriginal))
+      val writerUuid = values(6).getStringValue
+      val payloadAsBytes = Base64.getDecoder.decode(payloadAsString)
+      val payload = serialization.deserialize(payloadAsBytes, serId, serManifest).get
+      (
+        PersistentRepr(
+          payload,
+          sequenceNr,
+          persistenceId,
+          writerUuid = writerUuid
+        ).withTimestamp(writeTimestamp.toEpochMilli),
+        writeOriginal
+      )
+    }
   }
 
   val ReplayTypes = Map(
@@ -133,10 +168,12 @@ private[spanner] class SpannerInteractions(spannerGrpcClient: SpannerGrpcClient,
   }
 
   def readHighestSequenceNr(persistenceId: String, fromSequenceNr: Long): Future[Long] =
-    spannerGrpcClient.withSession(implicit session => internalReadHighestSequenceNr(persistenceId, fromSequenceNr))
+    spannerGrpcClient.withSession(
+      implicit session => internalReadHighestSequenceNr(persistenceId, fromSequenceNr)
+    )
 
-  def deleteMessagesTo(persistenceId: String, toSequenceNr: Long): Future[Unit] = spannerGrpcClient.withSession {
-    implicit session =>
+  def deleteMessagesTo(persistenceId: String, toSequenceNr: Long): Future[Unit] =
+    spannerGrpcClient.withSession { implicit session =>
       log.debug("deleteMessagesTo, session id [{}]", session.id)
 
       def params(to: Long) =
@@ -162,7 +199,7 @@ private[spanner] class SpannerInteractions(spannerGrpcClient: SpannerGrpcClient,
           )
         )
       } yield ()
-  }
+    }
 
   def streamJournal(
       serialization: Serialization,
@@ -172,22 +209,7 @@ private[spanner] class SpannerInteractions(spannerGrpcClient: SpannerGrpcClient,
       max: Long
   )(replay: PersistentRepr => Unit): Future[Unit] = {
     def replayRow(values: Seq[Value]): Unit = {
-      val persistenceId = values.head.getStringValue
-      val sequenceNr = values(1).getStringValue.toLong
-      val payloadAsString = values(2).getStringValue
-      val serId = values(3).getStringValue.toInt
-      val serManifest = values(4).getStringValue
-      val writerUuid = values(6).getStringValue
-
-      val payloadAsBytes = Base64.getDecoder.decode(payloadAsString)
-      val payload = serialization.deserialize(payloadAsBytes, serId, serManifest).get
-      val repr = PersistentRepr(
-        payload,
-        sequenceNr,
-        persistenceId,
-        writerUuid = writerUuid
-      )
-
+      val (repr, _) = Schema.Journal.deserializeRow(serialization, values)
       log.debug("replaying {}", repr)
       replay(repr)
     }
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
@@ -217,12 +217,14 @@ private[spanner] class SpannerInteractions(spannerGrpcClient: SpannerGrpcClient,
     spannerGrpcClient
       .streamingQuery(
         ReplaySql,
-        params = Struct(
-          fields = Map(
-            Journal.PersistenceId._1 -> Value(StringValue(persistenceId)),
-            "from_sequence_nr" -> Value(StringValue(fromSequenceNr.toString)),
-            "to_sequence_nr" -> Value(StringValue(toSequenceNr.toString)),
-            "max" -> Value(StringValue(max.toString))
+        params = Some(
+          Struct(
+            fields = Map(
+              Journal.PersistenceId._1 -> Value(StringValue(persistenceId)),
+              "from_sequence_nr" -> Value(StringValue(fromSequenceNr.toString)),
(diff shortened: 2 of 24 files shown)


chbatey:
Snuck in all persistence ids to try and re-use the continuous query

journal/src/main/resources/reference.conf
```diff
@@ -10,6 +10,10 @@ akka.persistence.spanner {
 
  query {
   class = "akka.persistence.spanner.SpannerReadJournalProvider"
+
+  # When live queries return no results. How often to poll Spanner for new
+  # rows
+  refresh-interval = 3s
  }
```
patriknw, line 17: indentation off by one

journal/src/main/scala/akka/persistence/spanner/SpannerOffset.scala
```diff
@@ -0,0 +1,16 @@
...
+package akka.persistence.spanner
+
+import akka.persistence.query.Offset
+
+/**
+ * @param commitTimestamp microsecond granularity stored in nanoseconds
+ * @param seen List of sequence nrs for every persistence id seen at this timestamp
+ */
+final case class SpannerOffset(
+    commitTimestamp: String,
```
patriknw, line 14: would it be better to use `java.sql.Timestamp` or is it anyway converted to a String for gRPC?

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
```diff
@@ -0,0 +1,144 @@
...
+          val source = nextQuery(state)
+          log.debug("Next source {}. Current state {} {}", source, nextRow, state)
+          source match {
+            case Some(source) =>
+              sinkIn = new SubSinkInlet[T]("Yep")
+              sinkIn.setHandler(new InHandler {
+                override def onPush(): Unit =
+                  if (isAvailable(out)) {
+                    if (!nextRow.isEmpty) {
+                      throw new RuntimeException(s"onPush called when we already have: " + nextRow)
```
patriknw, line 92: IllegalStateException

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
```diff
@@ -4,18 +4,168 @@
...
+      .streamingQuery(PersistenceIdsQuery)
+      .map { values =>
+        values.head.getStringValue
+      }
+      .mapMaterializedValue(_ => NotUsed)
+  }
+
+  override def persistenceIds(): Source[String, NotUsed] = {
+    log.debug("persistenceIds")
+    ContinuousQuery[Unit, String](
```
patriknw, line 154: nice

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
```diff
@@ -4,18 +4,168 @@
...
+                  pr.persistenceId,
+                  pr.sequenceNr,
+                  pr.payload,
+                  pr.timestamp
+                )
+              )
+            }
+          }
+        }
+      }
```
patriknw, line 116: looks simple, nice

journal/src/main/scala/akka/persistence/spanner/SpannerOffset.scala
chbatey: I was concerned about doing any parsing to and from string formats of the spanner commit write timestamp. It is received and sent as a string (protocol is limited to json types even though it is over protobuf). E.g. right now commit timestamps are given in nano seconds but only to microsecond granularity, it would be easy when turning it back to a string to mess something up if spanner changes this

journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
```diff
@@ -37,6 +39,39 @@ private[spanner] object Schema {
...
+      .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+      .optionalStart
+      .appendOffsetId
+      .optionalEnd
+      .toFormatter
+      .withZone(ZoneOffset.UTC)
+
+    val NoOffset = formatter.format(Instant.ofEpochMilli(0))
+
+    def deserializeRow(serialization: Serialization, values: Seq[Value]): (PersistentRepr, String) = {
```
johanandren, line 53: Nice to colocate things depending on schema

journal/src/main/scala/akka/persistence/spanner/SpannerOffset.scala
johanandren: What i've seen so far is that the spanner grpc string formats does not include micros granularity for TIMESTAMP columns, but maybe I did something wrong in how I specified the datatypes.

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -29,6 +29,10 @@ private[spanner] object SpannerSettings {
   }
 }
 
+private[spanner] final class QuerySettings(config: Config) {
+  val refreshInterval: FiniteDuration = config.getDuration("refresh-interval").asScala
+}
```
johanandren, line 34: I was lazy and didn't create a settings class with scope for snapshots but perhaps I should've

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
```diff
@@ -0,0 +1,144 @@
...
+            }
+          case OptionVal.None =>
+            if (!subStreamFinished && !sinkIn.isClosed && !sinkIn.hasBeenPulled)
+              sinkIn.pull()
+        }
+      }
+
+      setHandler(out, this)
+    }
+}
```
johanandren, line 144: Looks suspiciously simple!

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
```diff
@@ -0,0 +1,144 @@
...
+          scheduleOnce(NextQuery, refreshInterval)
+        } else {
+          log.debug("Running next query now")
+          nrElements = 0
+          subStreamFinished = false
+          val source = nextQuery(state)
+          log.debug("Next source {}. Current state {} {}", source, nextRow, state)
+          source match {
+            case Some(source) =>
+              sinkIn = new SubSinkInlet[T]("Yep")
```
johanandren, line 87: Doesn't this mean we could be overwriting a substream already running, we don't check `substreamFinished` anywher here afaics?

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
```diff
@@ -0,0 +1,144 @@
...
+      }
+
+      override protected def onTimer(timerKey: Any): Unit = timerKey match {
+        case NextQuery => next()
+      }
+
+      def next(): Unit =
+        if (nrElements <= threshold) {
+          log.debug("Scheduling next query for later")
+          nrElements = Long.MaxValue
```
johanandren, line 77: Hmmm, not sure I quite understand how the nrElements are used, i thought it is the number of elements that has passed through since last substream stream was started, but here we reset it?

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
```diff
@@ -0,0 +1,144 @@
...
+
+                override def onUpstreamFinish(): Unit =
+                  if (nextRow.isDefined) {
+                    log.debug("Stream finished. Not creating next as a buffered element")
+                    // wait for the element to be pulled
+                    subStreamFinished = true
+                  } else {
+                    log.debug("Stream finished. Creating next. " + nextRow)
+                    next()
+                  }
```
johanandren, line 111: onUpstreamFailure?

journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
```diff
@@ -37,6 +39,39 @@ private[spanner] object Schema {
...
+      val persistenceId = values.head.getStringValue
+      val sequenceNr = values(1).getStringValue.toLong
+      val payloadAsString = values(2).getStringValue
+      val serId = values(3).getStringValue.toInt
+      val serManifest = values(4).getStringValue
+      // keep this in the offset as the original format rather than do any conversions
+      val writeOriginal = values(5).getStringValue
```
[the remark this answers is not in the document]
chbatey: if the timestamp was written as a commit write timestamp they come back with microseconds. I originally had a bug when i started parsing it as I truncated it to milliseconds so thought it best to leave it exactly how spanner gives us for the offst

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
chbatey: we only call next() when the sub stream has finished. the `substreamFinished` var is for keeping track that the substream has finished but we have a row buffered

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
chbatey: You're right

Here we set to to MaxValue so when the timer call next() it will create the next stream

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
chbatey: The default impl seems sufficient? Fails the stage?

chbatey:
Test is flaky in travis so will run just that one for a while and see what's going on

journal/src/main/scala/akka/persistence/spanner/SpannerInteractions.scala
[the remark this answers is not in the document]
johanandren: Aha, that's the difference, for the snapshots I was the one writing the timestamp, got it.

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
johanandren: Is that _this_ stage? Always confusing with the substreams.

johanandren:
Sorry for the conflict.

journal/src/main/scala/akka/persistence/spanner/internal/ContinuousQuery.scala
chbatey: I added a tet, seems to work

johanandren: Thx

journal/src/main/resources/reference.conf
johanandren: #63

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -51,4 +55,6 @@ private[spanner] final class SpannerSettings(config: Config) {
   val snapshotsTable = config.getString("snapshot.table")
   val sessionPool = new SessionPoolSettings(config.getConfig("session-pool"))
   val sessionAcquisitionTimeout = config.getDuration("session-acquisition-timeout").asScala
+
+  val queryConfig = new QuerySettings(config.getConfig("query"))
```
johanandren, line 59: querySettings

chbatey: Will change this in my current work to avoid rebuild

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
```diff
@@ -4,18 +4,169 @@
...
+    with PersistenceIdsQuery {
+  private val log = LoggerFactory.getLogger(classOf[SpannerReadJournal])
   private val sharedConfigPath = cfgPath.replaceAll("""\.query$""", "")
+  private val settings = new SpannerSettings(system.settings.config.getConfig(sharedConfigPath))
+  private val serialization = SerializationExtension(system)
 
   private val grpcClient = SpannerGrpcClientExtension(system.toTyped).clientFor(sharedConfigPath)
+
+  private val EventsByTagSql =
+    s"SELECT ${Schema.Journal.Columns.mkString(",")} from ${settings.journalTable} WHERE @tag IN UNNEST(tags) AND write_time >= @write_time ORDER BY write_time, persistence_id, sequence_nr "
```
jroper, line 49: This isn't using an index on the tags field is it? (Last I looked, indexes on array fields in Spanner weren't supported). This has a few consequences that we perhaps should document:

* Scaling high throughput writes can't be achieved by sharding tags, since it doesn't remove the hotspot on the commit timestamp.
* If you have many different entity types (eg in Cloudstate we're considering having a single Spanner database per customer, so many services will use the same database), you must configure them to have different tables if you don't want their write throughput to impact each other.

PR #55, opened 2020-04-24
Snapshot store

johanandren:
References #41

### DIFF
--- a/journal/src/main/resources/reference.conf
+++ b/journal/src/main/resources/reference.conf
@@ -6,6 +6,12 @@ akka.persistence.spanner {
 
  journal {
   class = "akka.persistence.spanner.SpannerJournal"
+
+  # name of the table to use for events
+  table = "journal"
+
+  # name of the auxilary table used to store information about deletes
+  deletions-table = "deletions"
  }
 
  query {
@@ -13,14 +19,11 @@ akka.persistence.spanner {
  }
 
  snapshot {
-
+   class = "akka.persistence.spanner.SpannerSnapshotStore"
+   table = "snapshots"
  }
 
- # name of the table to use for events
- table = "journal"
 
- # name of the auxilary table used to store information about deletes
- deletions-table = "deletions"
 
  # config location for the grpc client.
  # located under akka.grpc.client
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
@@ -13,9 +13,9 @@ import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
 import akka.event.Logging
 import akka.persistence.journal.{AsyncWriteJournal, Tagged}
-import akka.persistence.spanner.SpannerInteractions.SerializedWrite
+import akka.persistence.spanner.internal.SpannerJournalInteractions.SerializedWrite
 import akka.persistence.spanner.SpannerJournal.WriteFinished
-import akka.persistence.spanner.internal.SpannerGrpcClientExtension
+import akka.persistence.spanner.internal.{SpannerGrpcClientExtension, SpannerJournalInteractions}
 import akka.persistence.{AtomicWrite, PersistentRepr}
 import akka.serialization.{Serialization, SerializationExtension, Serializers}
 import com.typesafe.config.Config
@@ -44,7 +44,7 @@ final class SpannerJournal(config: Config, cfgPath: String) extends AsyncWriteJo
 
   private val grpcClient = SpannerGrpcClientExtension(system.toTyped).clientFor(sharedConfigPath)
 
-  private val spannerInteractions = new SpannerInteractions(
+  private val spannerInteractions = new SpannerJournalInteractions(
     grpcClient,
     journalSettings
   )
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
@@ -129,8 +129,8 @@ private[spanner] final class SessionPool(
 
   override def onMessage(msg: Command): Behavior[Command] = msg match {
     case gt @ GetSession(replyTo, id) =>
-      if (log.isDebugEnabled()) {
-        log.debug(
+      if (log.isTraceEnabled()) {
+        log.trace(
           "GetSession from [{}], inUseSessions [{}], availableSessions [{}]",
           replyTo,
           inUseSessions.mkString(", "),
@@ -151,7 +151,7 @@ private[spanner] final class SessionPool(
       }
       this
     case ReleaseSession(id) =>
-      log.debug("ReleaseSession {}", id)
+      log.trace("ReleaseSession [{}]", id)
       if (inUseSessions.contains(id)) {
         val session = inUseSessions(id)
         inUseSessions -= id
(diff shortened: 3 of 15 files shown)


journal/src/main/resources/reference.conf
```diff
@@ -6,21 +6,24 @@ akka.persistence.spanner {
 
  journal {
   class = "akka.persistence.spanner.SpannerJournal"
+
+  # name of the table to use for events
+  table = "journal"
+
+  # name of the auxilary table used to store information about deletes
+  deletions-table = "deletions"
  }
```
patriknw, line 15: looks like indentation is off by one

journal/src/main/resources/reference.conf
```diff
@@ -6,21 +6,24 @@ akka.persistence.spanner {
 
  journal {
   class = "akka.persistence.spanner.SpannerJournal"
+
+  # name of the table to use for events
+  table = "journal"
```
chbatey, line 11: i had this at the upper layer as the query plugin uses it too, but also fine here

journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
```diff
@@ -44,7 +44,7 @@ final class SpannerJournal(config: Config, cfgPath: String) extends AsyncWriteJo
 
   private val grpcClient = SpannerGrpcClientExtension(system.toTyped).clientFor(sharedConfigPath)
 
-  private val spannerInteractions = new SpannerInteractions(
+  private val spannerInteractions = new SpannerJournalInteractions(
```
chbatey, line 47: 👍

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -151,7 +151,7 @@ private[spanner] final class SessionPool(
       }
       this
     case ReleaseSession(id) =>
-      log.debug("ReleaseSession {}", id)
+      log.trace("ReleaseSession [{}]", id)
```
chbatey, line 154: 👍 to trace

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSnapshotInteractions.scala
```diff
@@ -0,0 +1,243 @@
...
+  private val log = LoggerFactory.getLogger(getClass)
+
+  private val serialization: Serialization = SerializationExtension(system)
+  private val metaSerializer = serialization.serializerFor(classOf[SnapshotMetadata])
+
+  def findSnapshot(persistenceId: String, criteria: SnapshotSelectionCriteria): Future[Option[SelectedSnapshot]] = {
+    val query =
+      "SELECT ser_id, ser_manifest, snapshot, meta_ser_id, meta_ser_manifest, meta " +
+      s"FROM ${settings.snapshotsTable} " +
+      wherePartFor(criteria) +
```
chbatey, line 75: the four types could still be constants

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSnapshotInteractions.scala
```diff
@@ -0,0 +1,243 @@
...
+                queryParamTypes
+              )
+            )
+          )
+      }
+  }
+
+  private def wherePartFor(criteria: SnapshotSelectionCriteria): String =
+    "WHERE persistence_id = @persistence_id " +
+    (if (criteria.maxSequenceNr != Long.MaxValue) "AND sequence_nr <= @max_seq_nr " else "") +
```
chbatey, line 212: I wonder if we need any indexes to make these effecient

journal/src/test/resources/logback-test.xml
```diff
@@ -9,7 +9,7 @@
   </appender>
   <logger name="io.netty" level="ERROR" />
   <logger name="io.grpc" level="INFO" />
-  <root level="INFO">
+  <root level="TRACE">
```
chbatey, line 12: probably don't leave this on, travis gets upset about too much output

journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
```diff
@@ -121,6 +124,21 @@ object SpannerSpec {
         deleted_to INT64 NOT NULL,
       ) PRIMARY KEY (persistence_id)
   """
+
+  val snapshotTable =
+    """
+    CREATE TABLE snapshots (
+      persistence_id STRING(MAX) NOT NULL,
```
chbatey, line 131: looks good

journal/src/test/scala/akka/persistence/spanner/internal/SpannerUtilsSpec.scala
```diff
@@ -0,0 +1,13 @@
+package akka.persistence.spanner.internal
+
+import org.scalatest.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class SpannerUtilsSpec extends AnyWordSpec with Matchers {
+  "The Spanner utils" must {
+    "format unix timestamps as rfc3339" in {
+      val timestamp = 1587658225325L
+      SpannerUtils.spannerTimestamp(timestamp) should ===("2020-04-23T16:10:25.325Z")
```
chbatey, line 10: FYI they can be stored to the nano second granularity in this format

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSnapshotInteractions.scala
johanandren: Yep, that is likely. Haven't looked into that yet though.

journal/src/test/resources/logback-test.xml
johanandren: Set to DEBUG now, we should capture logs eventually.

journal/src/test/scala/akka/persistence/spanner/internal/SpannerUtilsSpec.scala
johanandren: I couldn't find anything else then that it needs to be represented with this RFC3339 format in queries and results, so that requires some trixery to actually interact with?

journal/src/main/resources/reference.conf
johanandren: Yeah, @chbatey lets reformat when we don't have so much work in progress.

journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
johanandren: Mimic:ed the couchbase plugin too much without thinking, the metadata does not need to be serialized and stored because all it contains are pid,seqnr and timestamp which were already in the table.

johanandren: Sorry, the couchbase plugin doesn't do that at all, I just made that up.

johanandren:
All tests passing now 🧇

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSnapshotInteractions.scala
johanandren: Thinking a bit more about it, pid and seqnr are primary key, and we always filter on pid first, having many snapshots for the same pid feels like a special case, so maybe that's rather up to the use case?

chbatey:
Headers, maybe we should add that to compile

Issue #42, opened 2020-04-21
Unknown sessions returned


johanandren:
Get a lot of these when the perfspec is failing: 

```
[...]

PR #58, opened 2020-04-24
Keep outstanding requests in queue instead of stashing messages

johanandren:
References #42

This allows us to remove queued requests if the requestee gives up before getting one

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
@@ -144,11 +144,11 @@ private[spanner] final class SessionPool(
       if (availableSessions.nonEmpty) {
         handOutSession(id, replyTo)
       } else {
-        if (availableSessions.size >= settings.maxOutstandingRequests) {
+        if (requestQueue.size >= settings.maxOutstandingRequests) {
           log.warn("Session pool request stash full, denying request for pool")
           replyTo ! PoolBusy(id)
         } else {
-          log.debug("No free sessions, enqueuing request for session [{}]", id)
+          log.trace("No free sessions, enqueuing request for session [{}]", id)
           requestQueue.enqueue(gt)
         }
       }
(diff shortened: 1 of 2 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -117,57 +117,62 @@ private[spanner] final class SessionPool(
...
       if (availableSessions.nonEmpty) {
-        val next = availableSessions.dequeue()
-        replyTo ! PooledSession(next.session, id)
-        inUseSessions += (id -> next.session)
+        handOutSession(id, replyTo)
       } else {
-        if (stash.isFull) {
-          ctx.log.warn("Session pool request stash full, denying request for pool")
+        if (availableSessions.size >= settings.maxOutstandingRequests) {
+          log.warn("Session pool request stash full, denying request for pool")
```
chbatey, line 144: requestQueue.size

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -117,57 +117,62 @@ private[spanner] final class SessionPool(
...
         availableSessions.enqueue(AvailableSession(session, System.currentTimeMillis()))
-        stash.unstash(this, 1, identity)
+        handOutSessionToQueuedRequest()
+      } else if (requestQueue.exists(_.id == id)) {
+        // requestor gave up before they got a session
+        log.trace("Queued session request [{}] given up without getting a session", id)
+        requestQueue = requestQueue.filterNot(_.id == id)
       } else {
-        log.error("unknown session returned [{}]. This is a bug.", id)
+        log.error("Unknown session returned [{}], this is a bug", id)
```
chbatey, line 164: i don't think this is a bug any more, it is expected if we timeout

chbatey: oh right this is a bug, nvm

chbatey:
ah sorry for the conflict

chbatey:
@johanandren i think it is worth getting this in before doing any testing

PR #67, opened 2020-04-29
Configuration for session stats logging

chbatey:
On top of #65


journal/src/main/resources/reference.conf
```diff
@@ -56,6 +59,13 @@ akka.persistence.spanner {
   # How often to send a dummy query for each session to keep it alive
   # Spanner drops sessions after 60 minutes
   keep-alive-interval = 25 minutes
+
+  # Internal between each log of stats
+  # set to off to disable
+  stats-interval = 3s
```
johanandren, line 65: off as default perhaps? or at least more seldom than every 3s

chbatey: Okay

chbatey: I'll do this in my next PR to avoid too many rebuilds

PR #69, opened 2020-04-29
End to end event by tag example

johanandren:
Based on the C* PR


johanandren:
Got stuck on no logs on run, leaving it for tomorrow.

johanandren:
If you write such instructions in the original PR I'll copy it here ;)

chbatey:
THe failure is events by persistence id spec which seems to be unlikely to this change

I'll merge this and investigate in master

PR #70, opened 2020-04-30
Session Pool shutdown

chbatey:
On top of #67 and #65


build.sbt
```diff
@@ -2,7 +2,6 @@ import sbt.Keys.parallelExecution
 
 inThisBuild(
   Seq(
-    scalaVersion := "2.13.1",
```
johanandren, line 5: Huh?

journal/src/main/resources/reference.conf
```diff
@@ -56,6 +59,16 @@ akka.persistence.spanner {
...
+
+  # Internal between each log of stats
+  # set to off to disable
+  stats-interval = 3s
+  # Stats logger to use, set this logger to DEBUG in your configuration
+  # to see the stats logging
+  stats-logger = "akka.persistence.spanner.SessionPool.stats"
+
+  # Keep lower than coordinated shutdown stage timeout
+  shutdown-timeout = 3s
```
johanandren, line 71: Also document what it times out in addition to the warning?

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -232,17 +319,23 @@ private[spanner] final class SessionPool(
       uses = 0
       this
 
+    case Shutdown(done) =>
+      timers.cancel(KeepAlive)
+      timers.cancel(Stats)
+      timers.startSingleTimer(ShutdownTimeout, settings.sessionPool.shutdownTimeout)
+      SessionPool.shuttingDown(done, client, availableSessions.toList.map(_.session), inUseSessions)
```
johanandren, line 326: If we have idle sessions, shouldn't we immediately trigger Delete of those here rather than passing to shutting down behavior?

chbatey: Yep we can do that

PR #79, opened 2020-05-06
Hash pid to tag and verify seqnr in sequence on read side in example

johanandren:


example/src/main/scala/akka/persistence/spanner/example/EventProcessorStream.scala
```diff
@@ -95,35 +85,54 @@ class EventProcessorStream[Event: ClassTag](
...
+          if (counter % 100 == 0) log.info("Successfully processed [{}] events for tag [{}]", counter, tag)
+          ee :: Nil
+        }
+      }
+      .map { eventEnvelope =>
+        eventEnvelope.event match {
+          case event: Event => {
+            // Times from different nodes, take with a pinch of salt
+            val latency = System.currentTimeMillis() - eventEnvelope.timestamp
+            if (latency < histogram.getMaxValue)
```
johanandren, line 112: I think this isn't working as expected, I see a lot of dropped values that are less than the expected minute, like sub second. Not sure what is going on.

Issue #77, opened 2020-05-04
Indexes for queries


chbatey:
Particularly the events by tag. Can we do it with arrays or do we need to change the schema.

johanandren:
I don't think we need to provide indexes for the other queries out of the box, some of them should  already be covered by the primary key, but users can also add indexes for the other queries if it turns they are important to an application and they are slow.

PR #80, opened 2020-05-06
Separated interleaved table for tags 

johanandren:
So that we can index tags. This is the needed schema change will look into secondary index for it next.

References #77

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
@@ -40,18 +40,16 @@ private[spanner] object SpannerJournalInteractions {
 
   object Schema {
     object Journal {
-      // weird formatting is for docs
       def journalTable(settings: SpannerSettings): String =
         s"""CREATE TABLE ${settings.journalTable} (
-        persistence_id STRING(MAX) NOT NULL,
-        sequence_nr INT64 NOT NULL,
-        event BYTES(MAX),
-        ser_id INT64 NOT NULL,
-        ser_manifest STRING(MAX) NOT NULL,
-        write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
-        writer_uuid STRING(MAX) NOT NULL,
-) PRIMARY KEY (persistence_id, sequence_nr)
-"""
+           |  persistence_id STRING(MAX) NOT NULL,
+           |  sequence_nr INT64 NOT NULL,
+           |  event BYTES(MAX),
+           |  ser_id INT64 NOT NULL,
+           |  ser_manifest STRING(MAX) NOT NULL,
+           |  write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
+           |  writer_uuid STRING(MAX) NOT NULL,
+           |) PRIMARY KEY (persistence_id, sequence_nr)""".stripMargin
 
       val PersistenceId = "persistence_id" -> Type(TypeCode.STRING)
       val SeqNr = "sequence_nr" -> Type(TypeCode.INT64)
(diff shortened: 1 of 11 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
```diff
@@ -73,52 +72,66 @@ private[spanner] object SpannerJournalInteractions {
...
         val payload = serialization.deserialize(payloadAsBytes, serId, serManifest).get
         (
-          PersistentRepr(
-            payload,
-            sequenceNr,
-            persistenceId,
-            writerUuid = writerUuid
-          ).withTimestamp(writeTimestamp.toEpochMilli),
+          PersistentRepr(payload, sequenceNr, persistenceId, writerUuid = writerUuid)
+            .withTimestamp(writeTimestamp),
```
johanandren, line 99: some unrelated restructure and small fixes here
1. re-use logic for spanner timeformat
2. don't index a seq we don't know is an IndexedSeq
3. move stuff related to Journal table into that object.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
```diff
@@ -73,52 +72,66 @@ private[spanner] object SpannerJournalInteractions {
...
+    object Tags {
+      def tagTable(settings: SpannerSettings): String =
+        // weird formatting is for docs
+        s"""CREATE TABLE ${settings.eventTagTable} (
+      persistence_id STRING(MAX) NOT NULL,
+      sequence_nr INT64 NOT NULL,
+      tag STRING(MAX) NOT NULL
+) PRIMARY KEY (persistence_id, sequence_nr, tag),
+  INTERLEAVE IN PARENT ${settings.journalTable}
+      """
```
johanandren, line 114: Interleaving it means it will always be stored together with the event, which is how we always read the tags.

journal/src/test/scala/akka/persistence/spanner/SpannerSpec.scala
```diff
@@ -203,12 +201,16 @@ trait SpannerLifecycle
...
-          SpannerJournalInteractions.Schema.deleteMetadataTable(spannerSettings) ::
+          SpannerJournalInteractions.Schema.Tags.tagTable(spannerSettings) ::
+          SpannerJournalInteractions.Schema.Deleted.deleteMetadataTable(spannerSettings) ::
           (if (withSnapshotStore)
              SpannerSnapshotInteractions.Schema.Snapshots.snapshotTable(spannerSettings) :: Nil
            else Nil)
         )
       )
+      .onFailure {
+        case ex => ex.printStackTrace()
```
johanandren, line 212: Without this I didn't get details about what was wrong with my incorrect table definition ;)

johanandren:
Verified this with the example against actual spanner and got up to 200k rows with the events by tag query staying at just above 10 ms. Not sure if I need to push that even further to really be sure it scales.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
```diff
@@ -73,52 +69,74 @@ private[spanner] object SpannerJournalInteractions {
...
+           |  sequence_nr INT64 NOT NULL,
+           |  tag STRING(MAX) NOT NULL,
+           |  write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
+           |) PRIMARY KEY (persistence_id, sequence_nr, tag),
+           |INTERLEAVE IN PARENT ${settings.journalTable}""".stripMargin
 
-    // weird formatting is for docs
-    def deleteMetadataTable(settings: SpannerSettings): String =
-      s"""CREATE TABLE ${settings.deletionsTable} (
+      def eventsByTagIndex(settings: SpannerSettings): String =
```
chbatey, line 113: do you see much difference for 200k events with and without the index?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
```diff
@@ -40,19 +40,16 @@ private[spanner] object SpannerJournalInteractions {
...
-        write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
-        writer_uuid STRING(MAX) NOT NULL,
-) PRIMARY KEY (persistence_id, sequence_nr)
-"""
+           |  persistence_id STRING(MAX) NOT NULL,
+           |  sequence_nr INT64 NOT NULL,
+           |  event BYTES(MAX),
+           |  ser_id INT64 NOT NULL,
+           |  ser_manifest STRING(MAX) NOT NULL,
+           |  write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
```
chbatey, line 50: do we need this in this table as well?

chbatey: Maybe we do, so that we can create offsets

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
johanandren: Yeap, that's 600+ ms

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
johanandren: Yes, it is needed so we can index the offset for the query. (I couldn't figure out a way to create an index for a join that contains values from both tables)

PR #89, opened 2020-05-13
Test with latest Akka gRPC snapshot

raboof:
Also tested on 1.8.0_252 (with the emulator)


chbatey:
Thanks @raboof! Should we add a cron to always test with the latest?

raboof:
Might be nice, but not sure if it's worth it - that might explode if we do it for all dependencies ;)

(superseded by #91)

PR #90, opened 2020-05-18
Testkit utility for creating and dropping database

chbatey:
I had planned to just document how to do this, and use the emutlator,
using the actual spanner client.

Unfortunately the spanner client can't be on the same classpath as APS
as it has a conficting version of grpc java.

Instead offering a very basic way to create and drop the schema via
our gRPC client.


testkit/src/main/resources/reference.conf
```diff
@@ -0,0 +1,4 @@
+akka.persistence.spanner.testkit {
+  # Timeout for operations such as creating the instance and database
+  operation-timeout = 5s
+}
```
patriknw, line 4: line break could be important here

testkit/src/main/scala/akka/persistence/spanner/testkit/SpannerTestkit.scala
```diff
@@ -0,0 +1,119 @@
...
+        ),
+      testkitSettings.operationTimeout
+    )
+  }
+
+  /**
+   * Drops the configured database.
+   */
+  def dropDatabase(): Unit =
+    Await.ready(
```
patriknw, line 115: intentionally not throwing if it timeout?

chbatey: Nope, thanks :)

PR #93, opened 2020-06-24
Akka serializer for the SpannerOffset

patriknw:
* needed for offset storage in Akka Projections

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+package akka.persistence.spanner.internal
+
+import java.io.NotSerializableException
+import java.nio.charset.StandardCharsets.UTF_8
+
+import scala.util.control.NonFatal
+
+import akka.actor.ExtendedActorSystem
+import akka.annotation.InternalApi
+import akka.persistence.spanner.SpannerOffset
+import akka.serialization.BaseSerializer
+import akka.serialization.SerializerWithStringManifest
+
+/**
+ * INTERNAL API: Serialized SpannerOffset is used by Akka Projections
+ */
+@InternalApi private[akka] class SpannerSerializer(val system: ExtendedActorSystem)
+    extends SerializerWithStringManifest
+    with BaseSerializer {
+  private val SpannerOffsetManifest = "a"
+
+  // persistenceId and commitTimestamp must not contain this separator char
+  private val separator = ';'
+
+  override def manifest(o: AnyRef): String = o match {
+    case _: SpannerOffset => SpannerOffsetManifest
+    case _ =>
+      throw new IllegalArgumentException(s"Can't serialize object of type ${o.getClass} in [${getClass.getName}]")
+  }
+
+  override def toBinary(o: AnyRef): Array[Byte] = o match {
+    case o: SpannerOffset => offsetToBinary(o)
+    case _ =>
+      throw new IllegalArgumentException(s"Cannot serialize object of type [${o.getClass.getName}]")
+  }
+
+  private def offsetToBinary(offset: SpannerOffset): Array[Byte] = {
+    val str = new java.lang.StringBuilder
+    str.append(offset.commitTimestamp)
+    if (offset.seen.size == 1) {
+      // optimized for the normal case
+      val pid = offset.seen.head._1
+      val seqNr = offset.seen.head._2
+      str.append(separator).append(pid).append(separator).append(seqNr)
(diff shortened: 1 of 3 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * INTERNAL API: Serialized SpannerOffset is used by Akka Projections
+ */
+@InternalApi private[akka] class SpannerSerializer(val system: ExtendedActorSystem)
+    extends SerializerWithStringManifest
+    with BaseSerializer {
+  private val SpannerOffsetManifest = "a"
+
+  // persistenceId and commitTimestamp must not contain this separator char
+  private val separator = ';'
```
patriknw, line 26: do you think there is a risk that we hit a collision for this char? anyone using it in the persistenceId?

journal/src/test/scala/akka/persistence/spanner/internal/SpannerSerializerSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+import akka.actor.ExtendedActorSystem
+import akka.actor.testkit.typed.scaladsl.LogCapturing
+import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
+import akka.persistence.spanner.SpannerOffset
+import akka.serialization.SerializationExtension
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class SpannerSerializerSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike with LogCapturing {
+  private val serializer = new SpannerSerializer(system.classicSystem.asInstanceOf[ExtendedActorSystem])
+  private val commitTimestamp = Timestamp.valueOf("2020-06-24 10:35:15.123456").toString
```
patriknw, line 17: the format of the timestamp doesn't really matter as long as it doesn't contain the separator char, but do you know what it looks like for real?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
```diff
@@ -0,0 +1,93 @@
...
+      )
+  }
+
+  private def offsetFromBinary(bytes: Array[Byte]): SpannerOffset = {
+    val str = new String(bytes, UTF_8)
+    try {
+      val parts = str.split(separator)
+      if (parts.length == 3) {
+        // optimized for the normal case
+        SpannerOffset(parts(0), Map(parts(1) -> parts(2).toLong))
```
patriknw, line 71: is my assumption right that the normal case will be 1 pid/seqNr?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
johanandren: Perhaps a more unusual separator char would be good. I suggest 〷 (telegraphy line separator) ;)

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
johanandren: I think it only ever has seen entries when two writes were at the same timestamp. But I'm not 100% sure I remember right.

journal/src/test/scala/akka/persistence/spanner/internal/SpannerSerializerSpec.scala
johanandren: It's rfc3339, you can use `SpannerUtils.*` to format

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
patriknw: at least `;` is escaped with URLEncoder if someone for strange reason is url-encoding their persistenceIds

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSerializer.scala
chbatey: @patriknw is right. There's always at least one, the event found at this timestamp. So when we start from that timestamp again it is used to filter out the event from last time.

chbatey: And yes it is nearly always only 1, unless multiple were written in the same microsecond

patriknw:
@hseeberger Let us know if you start using Akka Projections with Spanner. Then you need a new version of akka-persistence-spanner, that is including this serializer.

hseeberger:
Thanks, @patriknw. Currently we are planing to use Akka Projections for the management cluster only where we are using Postgres as our database.

PR #94, opened 2020-07-15
Akka gRPC 1.0

chbatey:


chbatey:
Needs an akka upgrade also

ignasi35:
> Needs an akka upgrade also

Worth a mention on the release notes.

PR #95, opened 2020-07-24
Replicated Event Sourcing support

johanandren:
Won't work until we release Akka with Active Active APIs.


journal/src/main/resources/reference.conf
```diff
@@ -4,6 +4,11 @@ akka.persistence.spanner {
   instance = "akka"
   database = "akka"
 
+  # Needs to be enabled to use the spanner plugin with active active entities, note that this will require
+  # additional table columns for replicated metadata that must be present in the schema and that a journal
+  # with replication meta enabled cannot be used for regular event sourced actors.
+  with-replication-meta = false
```
chbatey, line 10: maybe just with meta in case we use it for something else

journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
```diff
@@ -73,15 +74,35 @@ final class SpannerJournal(config: Config, cfgPath: String) extends AsyncWriteJo
...
+                val m = replicatedMeta.asInstanceOf[AnyRef]
+                val serializedMeta = serialization.serialize(m).get
+                val serializedMetaAsString = Base64.getEncoder.encodeToString(serializedMeta)
+                val metaSerializer = serialization.findSerializerFor(m)
+                val metaManifest = Serializers.manifestFor(metaSerializer, m)
+                val id: Int = metaSerializer.identifier
+                write.copy(metadata = Some(SerializedEventMetadata(id, metaManifest, serializedMetaAsString)))
+              case None =>
+                throw new IllegalArgumentException(
+                  s"with-replication-meta enabled but trying to write event (persistence id ${pr.persistenceId}, sequence nr ${pr.sequenceNr}) without metadata, " +
```
chbatey, line 101: why can't we mix? i'd imagine having some replicated ones and some not isn't that uncommon? Would we advise separate journals for that?

johanandren: Yeah, so we keep the toggle, but accept empty meta columns when it is enabled. I'll change it.

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
```diff
@@ -138,7 +142,7 @@ final class SpannerReadJournal(system: ExtendedActorSystem, config: Config, cfgP
       toSequenceNr: Long
   ): Source[EventEnvelope, NotUsed] =
     ContinuousQuery[Long, EventEnvelope](
-      fromSequenceNr,
+      fromSequenceNr - 1, // we always add 1 back below before querying
```
johanandren, line 145: Oh, actually, we don't if it is == toSequenceNr, so should probably fix that (even if it is unlikely to query for a single specific event)

journal/src/main/resources/reference.conf
```diff
@@ -4,6 +4,11 @@ akka.persistence.spanner {
   instance = "akka"
   database = "akka"
 
+  # Needs to be enabled to use the spanner plugin with active active entities, note that this will require
```
patriknw, line 7:
```suggestion
  # Needs to be enabled to use the spanner plugin with Replicated Event Sourced entities, note that this will require
```

journal/src/main/resources/reference.conf
```diff
@@ -4,6 +4,11 @@ akka.persistence.spanner {
...
+  # additional table columns for replicated metadata that must be present in the schema and that a journal
+  # with replication meta enabled cannot be used for regular event sourced actors.
```
patriknw, line 9: what is the reason for "cannot be used for regular event sourced actors" ?

journal/src/test/scala/akka/persistence/spanner/ActiveActiveSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+import akka.persistence.spanner.ActiveActiveSpec.MyActiveActiveStringSet
+import akka.persistence.spanner.javadsl.SpannerReadJournal
+import akka.persistence.typed.ReplicaId
+import akka.persistence.typed.scaladsl.ActiveActiveEventSourcing
+import akka.persistence.typed.scaladsl.Effect
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+
+import scala.concurrent.duration._
+
+object ActiveActiveSpec {
```
patriknw, line 15: Replicated...

project/Dependencies.scala
```diff
@@ -8,7 +8,8 @@ object Dependencies {
   val Scala212 = "2.12.11"
   val Scala213 = "2.13.1"
 
-  val AkkaVersion = System.getProperty("override.akka.version", "2.6.6")
+  // FIXME actual patch release with active active metadata
```
patriknw, line 11: replicated metadata

journal/src/main/resources/reference.conf
johanandren: I think I went back on that after discussing with Chris but missed removing this.

renatocaval:
closing and reopening to test the CLA validator

johanandren:
Ready for final review now

journal/src/main/resources/reference.conf
```diff
@@ -4,6 +4,11 @@ akka.persistence.spanner {
   instance = "akka"
   database = "akka"
 
+  # Needs to be enabled to use the spanner plugin with replicated eventsourcing,
+  # note that this will also require additional table columns for replicated metadata
+  # that must be present in the schema
+  with-meta = false
```
patriknw, line 10: could we enable this by default since there are probably no users of the plugin yet and we haven't released 1.0.0 final?

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
patriknw, line 145: is this an unrelated bug fix?

patriknw (commented):
looking good but I wonder if we can simplify by not making it an optional feature?

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
chbatey: It makes the replication streams work so unrelated but required

build.sbt
```diff
@@ -21,7 +21,11 @@ inThisBuild(
     licenses := Seq(("Apache-2.0", url("http://www.apache.org/licenses/LICENSE-2.0"))),
     description := "A replicated Akka Persistence journal backed by Spanner",
     // due to the emulator
-    parallelExecution := false
+    parallelExecution := false,
+    // add snapshot repo when Akka version overriden
+    resolvers ++=
+      (if (System.getProperty("override.akka.version") != null) Seq("Akka Snapshots".at("https://repo.akka.io/snapshots/"))
```
chbatey, line 27: Can remove this before merging

journal/src/main/resources/reference.conf
chbatey: 👍 i think this is a good idea given we hadn't done 1.0 yet

chbatey (approved):

journal/src/main/resources/reference.conf
johanandren: I'll remove the setting and make the meta columns mandatory 👍

journal/src/main/scala/akka/persistence/spanner/scaladsl/SpannerReadJournal.scala
johanandren: Yeah, noticed when adding this but broken before.

build.sbt
johanandren: I think it's good to keep to easily build/test against snaps. Doesn't hurt when not overriding Akka version either.

patriknw: I agree, it's a good idea that we should use in more places when needed

patriknw (approved):
LGTM

Issue #99, opened 2021-01-13
Session not discarded when request failing because spanner deleted it


johanandren:
In some scenario the logic in place to catch `io.grpc.StatusRuntimeException: NOT_FOUND: Session not found` and inform the pool that the session id is no longer valid, to trigger requesting a new session for the pool, does not work as expected.

Looks like it was observed both during event replay and persisting events.

johanandren:
Did some investigation and this is only happening during replay, which means only when doing the query through `SpannerGrpcClient#streamingQuery`. It tries to capture failed queries with a `.watchTermination` to make sure the session is released but perhaps the exception is wrapped or does not surface as expected failure of the stream itself.

johanandren:
The plot thickens:

The initial session failure is in fact an event write. `Failed to persist event type [Something] with sequence number [208943] for persistenceId [some-id]`.

There is one initial `User session failed with possible session not found. Re-creating session. ...`, which means the deleted session id is noticed correctly signalled back to the pool as no longer valid. 

What then follows is an large number of replay tries using that same session id, but at a somewhat lower sequence number `208940`. I initially thought that was the `streamingQuery` not handling the `NOT_FOUND` correctly but each of the log entires shows that the exception is thrown by Akka gRPC `UnaryCallAdapter.onClose` which means it is a call to a gRPC method with a single response.

So, two parts to the mystery:

* why does not the initial noticed `NOT_FOUND` lead to trashing the session id in the pool
* what single response operation is done during replay that is not covered with the session `NOT_FOUND` handling

PR #100, opened 2021-01-14
Failing write causing actor to get stuck

johanandren:
References #99

Turns out there was nothing specifically related to session invalidation, but rather that we forgot to clean out the write in progress map on write completion. This meant that a Future result of a failed write (say with the deleted session exception) would stay around forever, even though the error only happened once, and be returned from `asyncReadHighestSequenceNr` on every try to replay after that. 🤦

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
@@ -141,7 +141,8 @@ final class SpannerJournal(config: Config, cfgPath: String) extends AsyncWriteJo
     val pendingWrite = Option(writesInProgress.get(persistenceId)) match {
       case Some(f) =>
         log.debug("Write in progress for {}, deferring highest seq nr until write completed", persistenceId)
-        f
+        // we only want to make write - replay sequential, not fail if previous write failed
+        f.recover { case _ => Done }(ExecutionContexts.parasitic)
       case None => Future.successful(Done)
     }
     pendingWrite.flatMap(_ => spannerInteractions.readHighestSequenceNr(persistenceId, fromSequenceNr))
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
@@ -25,6 +25,10 @@ import scala.collection.immutable
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
 object SpannerJournal {
   case class WriteFinished(persistenceId: String, done: Future[_])
 }
@@ -110,8 +114,12 @@ final class SpannerJournal(config: Config, cfgPath: String) extends AsyncWriteJo
       }
     }
 
+    val persistenceId = messages.head.persistenceId
     val write = Future.sequence(messages.map(aw => atomicWrite(aw)))
-    writesInProgress.put(messages.head.persistenceId, write)
+    writesInProgress.put(persistenceId, write)
+    write.onComplete { _ =>
+      context.self ! WriteFinished(persistenceId, write)
+    }
     write
   }
 
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
@@ -219,7 +219,7 @@ private[spanner] final class SessionPool(
           handOutSessionToQueuedRequest()
         } else {
           log.debugN(
-            "Session released with should re-create. Not returning this session to the pool [{}] (all [{}])",
+            "Session released with should re-create. Not returning this session to the pool [{}]",
             session.name
           )
           ctx.self ! CreateSingleSession
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
@@ -218,7 +218,10 @@ private[spanner] final class SessionPool(
           availableSessions.enqueue(AvailableSession(session, System.currentTimeMillis()))
           handOutSessionToQueuedRequest()
         } else {
-          log.debug("Session released with should re-create. Not returning this session to the pool")
+          log.debugN(
+            "Session released with should re-create. Not returning this session to the pool [{}] (all [{}])",
+            session.name
+          )
           ctx.self ! CreateSingleSession
         }
       } else if (requestQueue.exists(_.id == id)) {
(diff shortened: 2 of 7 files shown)


journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
```diff
@@ -110,8 +114,12 @@ final class SpannerJournal(config: Config, cfgPath: String) extends AsyncWriteJo
       }
     }
 
+    val persistenceId = messages.head.persistenceId
     val write = Future.sequence(messages.map(aw => atomicWrite(aw)))
-    writesInProgress.put(messages.head.persistenceId, write)
+    writesInProgress.put(persistenceId, write)
+    write.onComplete { _ =>
+      context.self ! WriteFinished(persistenceId, write)
+    }
```
johanandren, line 122: This is the actual fix

journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
```diff
@@ -52,7 +52,7 @@ private[spanner] object SpannerGrpcClient {
  */
 @InternalApi private[spanner] class SpannerGrpcClient(
     name: String,
-    client: SpannerClient,
+    val client: SpannerClient,
```
johanandren, line 55: Needed access from the test to delete a session from spanner without the pool knowing

journal/src/test/scala/akka/persistence/spanner/SessionFaultToleranceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+
+      // replay fails
+      val persisterIncarnation2 = testKit.spawn(Persister(pid))
+      probe.expectTerminated(persisterIncarnation2)
+
+      // but session should be is invalidated, and a new one created so that the next operation succeeds
+      val persisterIncarnation3 = testKit.spawn(Persister(pid))
+      persisterIncarnation3 ! Persister.Ping(probe.ref)
+      probe.expectMessage(Done)
+    }
```
johanandren, line 43: This scenario never failed, just added for good measure.

journal/src/test/scala/akka/persistence/spanner/SessionFaultToleranceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+import com.typesafe.config.ConfigFactory
+
+class SessionFaultToleranceSpec extends SpannerSpec {
+  override def customConfig: Config = ConfigFactory.parseString("""
+       # single session makes testing easier
+       akka.persistence.spanner.session-pool.max-size = 1
+      """)
+
+  "Failure handling when the session is invalidated" must {
+    "recover from a replay failure" in {
```
johanandren, line 17: This scenario never failed, just added for good measure.

journal/src/test/scala/akka/persistence/spanner/TestActors.scala
```diff
@@ -31,11 +32,23 @@ object TestActors {
   }
 
   object Persister {
-    case class PersistMe(payload: Any, replyTo: ActorRef[Done])
+    sealed trait Command
+    case class PersistMe(payload: Any, replyTo: ActorRef[Done]) extends Command
+    case class Ping(replyTo: ActorRef[Done]) extends Command
```
johanandren, line 37: Just added a command to verify actor liveness here.

journal/src/test/scala/akka/persistence/spanner/internal/StubbedSpannerGrpcClientSpec.scala
```diff
@@ -118,7 +121,7 @@ class StubbedSpannerGrpcClientSpec
       val fakeClient = new AbstractStubbedSpannerClient {
         override def executeSql(in: ExecuteSqlRequest): Future[ResultSet] =
           Future.failed(
-            new StatusRuntimeException(Status.fromCode(Status.Code.NOT_FOUND).withDescription("Session not found"))
+            Status.fromCode(Status.Code.NOT_FOUND).withDescription("Session not found").asRuntimeException()
```
johanandren, line 124: Does not change anything but this is how Akka gRPC creates that exception.

journal/src/test/scala/akka/persistence/spanner/internal/StubbedSpannerGrpcClientSpec.scala
```diff
@@ -129,5 +132,24 @@ class StubbedSpannerGrpcClientSpec
       request.replyTo ! PooledSession(Session("cat"), request.id)
       pool.expectMessage(SessionPool.ReleaseSession(request.id, recreate = true))
     }
+
+    "request session to re-recreated if session NOT_FOUND during streaming query" in {
```
johanandren, line 136: Red herring for the failure but good to keep the extra coverage since the handling is a little bit different (stream watchTermination for noticing the failed query)

journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
johanandren: I wonder if we should perhaps recover the `write` future so that it never fails. We only have it to make write followed by replay sequential, not fail the following replay as now (we `flatMap` highest sequence nr query on top of it).

chbatey: ah i see, so we'd end up with a race that could cause the following reply to fail. Maybe the recover should be in the replay parts to be explicit that it doesn't care about an old write failing

chbatey (approved):
LGTM with or without the recover

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
```diff
@@ -218,7 +218,10 @@ private[spanner] final class SessionPool(
           availableSessions.enqueue(AvailableSession(session, System.currentTimeMillis()))
           handOutSessionToQueuedRequest()
         } else {
-          log.debug("Session released with should re-create. Not returning this session to the pool")
+          log.debugN(
+            "Session released with should re-create. Not returning this session to the pool [{}] (all [{}])",
+            session.name
```
patriknw, line 223: missing log args?

patriknw (approved):
LGTM, after fixing the log

journal/src/main/scala/akka/persistence/spanner/SpannerJournal.scala
johanandren: Yes, if there is a replay triggered before the failing write has been removed from `writesInProgress`, probably a very tiny time window for that but anyways. Recover in the replay side would avoid the extra work if not needed so that sounds good.

journal/src/main/scala/akka/persistence/spanner/internal/SessionPool.scala
johanandren: Ah, left over from trying to debug, all shouldn't be in there I'll remove that.

Issue #102, opened 2021-02-10
upsert/get/delete operations


raboof:
As underpinning for Durable Actors and cloudstate we expect to add upsert/get/delete operations to the Akka Persistence Plugin API. Let's start with some basic low-level implementations.

raboof:
done in #103

PR #103, opened 2021-02-11
Basic k/v table implementation

raboof:
Refs #102

### DIFF
--- a/journal/src/main/resources/reference.conf
+++ b/journal/src/main/resources/reference.conf
@@ -32,7 +32,9 @@ akka.persistence.spanner {
     table = "snapshots"
   }
 
-
+  object {
+    table = "objects"
+  }
 
   # config location for the grpc client.
   # located under akka.grpc.client
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -10,6 +10,7 @@ import akka.annotation.InternalStableApi
 import akka.persistence.spanner.internal.{SpannerGrpcClientExtension, SpannerObjectInteractions}
 import akka.util.ByteString
 
+import scala.util.control.NoStackTrace
 import scala.concurrent.Future
 
 /**
@@ -41,7 +42,7 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
  * INTERNAL API
  */
 object SpannerObjectStore {
-  case class ObjectNotFound(key: String) extends Exception(s"No data found for key [$key]")
+  case class ObjectNotFound(key: String) extends RuntimeException(s"No data found for key [$key]") with NoStackTrace
 
   case class Result(byteString: ByteString, serId: Long, serManifest: String, seqNr: Long)
 
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -8,9 +8,9 @@ import akka.actor.ActorSystem
 import akka.actor.typed.scaladsl.adapter.ClassicActorSystemOps
 import akka.annotation.InternalStableApi
 import akka.persistence.spanner.internal.{SpannerGrpcClientExtension, SpannerObjectInteractions}
+import akka.persistence.typed.PersistenceId
 import akka.util.ByteString
 
-import scala.util.control.NoStackTrace
 import scala.concurrent.Future
 
 /**
@@ -24,16 +24,16 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
 
   def upsertObject(
       entityType: String,
-      persistenceId: String,
+      persistenceId: PersistenceId,
       serId: Long,
       serManifest: String,
       value: ByteString,
       seqNr: Long
   ): Future[Unit] = interactions.upsertObject(entityType, persistenceId, serId, serManifest, value, seqNr)
 
-  def getObject(persistenceId: String): Future[Option[Result]] = interactions.getObject(persistenceId)
+  def getObject(persistenceId: PersistenceId): Future[Option[Result]] = interactions.getObject(persistenceId)
 
-  def deleteObject(persistenceId: String): Future[Unit] = interactions.deleteObject(persistenceId)
+  def deleteObject(persistenceId: PersistenceId): Future[Unit] = interactions.deleteObject(persistenceId)
 }
 
 /**
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -23,17 +23,17 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
   import SpannerObjectStore._
 
   def upsertObject(
-      entity: String,
-      key: String,
+      entityType: String,
+      persistenceId: String,
       serId: Long,
       serManifest: String,
       value: ByteString,
       seqNr: Long
-  ): Future[Unit] = interactions.upsertObject(entity, key, serId, serManifest, value, seqNr)
+  ): Future[Unit] = interactions.upsertObject(entityType, persistenceId, serId, serManifest, value, seqNr)
 
-  def getObject(key: String): Future[Option[Result]] = interactions.getObject(key)
+  def getObject(persistenceId: String): Future[Option[Result]] = interactions.getObject(persistenceId)
 
-  def deleteObject(key: String): Future[Unit] = interactions.deleteObject(key)
+  def deleteObject(persistenceId: String): Future[Unit] = interactions.deleteObject(persistenceId)
 }
 
 /**
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -31,7 +31,7 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
       seqNr: Long
   ): Future[Unit] = interactions.upsertObject(entity, key, serId, serManifest, value, seqNr)
 
-  def getObject(key: String): Future[Result] = interactions.getObject(key)
+  def getObject(key: String): Future[Option[Result]] = interactions.getObject(key)
 
   def deleteObject(key: String): Future[Unit] = interactions.deleteObject(key)
 }
@@ -42,8 +42,6 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
  * INTERNAL API
  */
 object SpannerObjectStore {
-  case class ObjectNotFound(key: String) extends RuntimeException(s"No data found for key [$key]") with NoStackTrace
-
   case class Result(byteString: ByteString, serId: Long, serManifest: String, seqNr: Long)
 
   def apply()(implicit system: ActorSystem): SpannerObjectStore = {
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.persistence.spanner
+
+import akka.actor.ActorSystem
+import akka.actor.typed.scaladsl.adapter.ClassicActorSystemOps
+import akka.annotation.InternalStableApi
+import akka.persistence.spanner.internal.{SpannerGrpcClientExtension, SpannerObjectInteractions}
+import akka.util.ByteString
+
+import scala.concurrent.Future
+
+/**
+ * This API would eventually evolve into the akka-persistence object API
+ *
+ * INTERNAL API
+ */
+@InternalStableApi
+class SpannerObjectStore(interactions: SpannerObjectInteractions) {
+  import SpannerObjectStore._
+
+  def upsertObject(
+      entity: String,
+      key: String,
+      serId: Long,
+      serManifest: String,
+      value: ByteString,
+      seqNr: Long
+  ): Future[Unit] = interactions.upsertObject(entity, key, serId, serManifest, value, seqNr)
+
+  def getObject(key: String): Future[Result] = interactions.getObject(key)
+
+  def deleteObject(key: String): Future[Unit] = interactions.deleteObject(key)
+}
+
+/**
+ * This API would eventually evolve into the akka-persistence object API
+ *
+ * INTERNAL API
+ */
+object SpannerObjectStore {
+  case class ObjectNotFound(key: String) extends Exception(s"No data found for key [$key]")
+
+  case class Result(byteString: ByteString, serId: Long, serManifest: String, seqNr: Long)
+
+  def apply()(implicit system: ActorSystem): SpannerObjectStore = {
+    val spannerSettings = new SpannerSettings(system.settings.config.getConfig("akka.persistence.spanner"))
+    val grpcClient = SpannerGrpcClientExtension(system.toTyped).clientFor("akka.persistence.spanner")
+    implicit val ec = system.dispatcher
+    val spannerInteractions = new SpannerObjectInteractions(
+      grpcClient,
+      spannerSettings
+    )
+    new SpannerObjectStore(spannerInteractions)
+  }
+}
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -4,7 +4,7 @@
 
 package akka.persistence.spanner
 
-import akka.annotation.InternalApi
+import akka.annotation.{InternalApi, InternalStableApi}
 import akka.persistence.spanner.SpannerSettings.SessionPoolSettings
 import com.typesafe.config.Config
 import akka.util.JavaDurationConverters._
@@ -44,7 +44,7 @@ private[spanner] final class QuerySettings(config: Config) {
 /**
  * INTERNAL API
  */
-@InternalApi
+@InternalStableApi
 final class SpannerSettings(config: Config) {
   val project = config.getString("project")
   val instance = config.getString("instance")
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -62,6 +62,10 @@ private[spanner] final class SpannerSettings(config: Config) {
   val maxWriteRetryTimeout = config.getDuration("max-write-retry-timeout").asScala
 
   val snapshotsTable = config.getString("snapshot.table")
+
+  // Object store to (eventually) back durable actors
+  val objectTable = config.getString("object.table")
+
   val sessionPool = new SessionPoolSettings(config.getConfig("session-pool"))
   val sessionAcquisitionTimeout = config.getDuration("session-acquisition-timeout").asScala
 
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -4,7 +4,7 @@
 
 package akka.persistence.spanner
 
-import akka.annotation.{InternalApi, InternalStableApi}
+import akka.annotation.InternalApi
 import akka.persistence.spanner.SpannerSettings.SessionPoolSettings
 import com.typesafe.config.Config
 import akka.util.JavaDurationConverters._
@@ -44,8 +44,8 @@ private[spanner] final class QuerySettings(config: Config) {
 /**
  * INTERNAL API
  */
-@InternalStableApi
-final class SpannerSettings(config: Config) {
+@InternalApi
+private[spanner] final class SpannerSettings(config: Config) {
   val project = config.getString("project")
   val instance = config.getString("instance")
   val database = config.getString("database")
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -45,7 +45,7 @@ private[spanner] final class QuerySettings(config: Config) {
  * INTERNAL API
  */
 @InternalApi
-private[spanner] final class SpannerSettings(config: Config) {
+final class SpannerSettings(config: Config) {
   val project = config.getString("project")
   val instance = config.getString("instance")
   val database = config.getString("database")
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
@@ -50,7 +50,7 @@ private[spanner] object SpannerGrpcClient {
 /**
  * A thin wrapper around the gRPC client to expose only what the plugin needs.
  */
-@InternalApi private[spanner] class SpannerGrpcClient(
+@InternalApi class SpannerGrpcClient(
     name: String,
     val client: SpannerClient,
     system: ActorSystem[_],
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.atomic.AtomicLong
 import akka.actor.typed.scaladsl.AskPattern._
 import akka.actor.typed.scaladsl.{Behaviors, LoggerOps}
 import akka.actor.typed.{ActorRef, ActorSystem, SupervisorStrategy}
-import akka.annotation.InternalApi
+import akka.annotation.{InternalApi, InternalStableApi}
 import akka.dispatch.ExecutionContexts
 import akka.persistence.spanner.SpannerSettings
 import akka.persistence.spanner.internal.SessionPool._
@@ -49,8 +49,10 @@ private[spanner] object SpannerGrpcClient {
 
 /**
  * A thin wrapper around the gRPC client to expose only what the plugin needs.
+ *
+ * INTERNAL API
  */
-@InternalApi class SpannerGrpcClient(
+@InternalStableApi class SpannerGrpcClient(
     name: String,
     val client: SpannerClient,
     system: ActorSystem[_],
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
@@ -121,7 +121,7 @@ private[spanner] object SpannerGrpcClient {
    */
   def executeBatchDml(statements: List[(String, Struct, Map[String, Type])])(
       implicit session: PooledSession
-  ): Future[Unit] =
+  ): Future[ExecuteBatchDmlResponse] =
     withWriteRetries { () =>
       def createBatchDmlRequest(sessionId: String, transactionId: ByteString): ExecuteBatchDmlRequest = {
         val s = statements.map {
@@ -168,7 +168,7 @@ private[spanner] object SpannerGrpcClient {
         )
       } yield {
         log.trace("Successful commit at {}", commitResponse.commitTimestamp.map(t => (t.seconds, t.nanos)))
-        ()
+        resultSet
       }
     }
 
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.atomic.AtomicLong
 import akka.actor.typed.scaladsl.AskPattern._
 import akka.actor.typed.scaladsl.{Behaviors, LoggerOps}
 import akka.actor.typed.{ActorRef, ActorSystem, SupervisorStrategy}
-import akka.annotation.{InternalApi, InternalStableApi}
+import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
 import akka.persistence.spanner.SpannerSettings
 import akka.persistence.spanner.internal.SessionPool._
@@ -49,10 +49,8 @@ private[spanner] object SpannerGrpcClient {
 
 /**
  * A thin wrapper around the gRPC client to expose only what the plugin needs.
- *
- * INTERNAL API
  */
-@InternalStableApi class SpannerGrpcClient(
+@InternalApi private[spanner] class SpannerGrpcClient(
     name: String,
     val client: SpannerClient,
     system: ActorSystem[_],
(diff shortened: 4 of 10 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,149 @@
...
+  object Schema {
+    // To back Durable Actors
+    object Objects {
+      def objectTable(settings: SpannerSettings): String =
+        s"""CREATE TABLE ${settings.objectTable} (
+           |  key STRING(MAX) NOT NULL,
+           |  value BYTES(MAX),
+           |  ser_manifest STRING(MAX) NOT NULL,
+           |  write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
+           |) PRIMARY KEY (key)""".stripMargin
```
johanandren, line 32: missing serializer id

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,157 @@
...
+
+      // TODO should we split the key into a persistenceId and an entityId?
+      val Key = "key" -> TypeCode.STRING
+      val Value = "value" -> TypeCode.BYTES
+      val SerId = "ser_id" -> TypeCode.INT64
+      val SerManifest = "ser_manifest" -> TypeCode.STRING
+      val WriteTime = "write_time" -> TypeCode.TIMESTAMP
+      // TODO consider adding a writerUUID like for eventsourced events?
+      // val WriterUUID = "writer_uuid" -> Type(TypeCode.STRING)
```
johanandren, line 42: Isn't that only interesting in when there is a sequence number that gets duplicated?

Optimistic locking could perhaps fill something like the same purpose here? A column with a counter for update and then look at if it has the right value to detect concurrent writers.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  private def wrapValue(column: (String, TypeCode), value: Any): Value =
+    Value((column, value) match {
+      case ((_, TypeCode.STRING), stringValue: String) => StringValue(stringValue)
+      case ((_, TypeCode.INT64), longValue: Long) => StringValue(longValue.toString)
+      case ((_, TypeCode.BYTES), bytes: ByteString) =>
+        // TODO verify why we're sending bytes as base64-encoded strings
+        // there might be some spanner restriction as the API used to be json/http rather than protobuf,
+        // but good to verify
+        StringValue(bytes.encodeBase64.utf8String)
```
johanandren, line 146: I think there is no way to send actual bytes with the spanner api but I may be wrong, perhaps @chbatey remembers better than me.

journal/src/main/resources/reference.conf
```diff
@@ -32,7 +32,9 @@ akka.persistence.spanner {
     table = "snapshots"
   }
 
-
+  object {
+    table = "objects"
```
patriknw, line 36: We might want to come up with a better name here. Current candidate in Akka is DurableStateBehavior so maybe `state` would be be better?

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -45,7 +45,7 @@ private[spanner] final class QuerySettings(config: Config) {
  * INTERNAL API
  */
 @InternalApi
```
patriknw, line 47: `InternalStableApi` ?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,157 @@
...
+ * INTERNAL API
+ */
+@InternalApi
+private[spanner] object SpannerObjectInteractions {
+  object Schema {
+    // To back Durable Actors
+    object Objects {
+      def objectTable(settings: SpannerSettings): String =
+        s"""CREATE TABLE ${settings.objectTable} (
+           |  key STRING(MAX) NOT NULL,
```
patriknw, line 28: alternative would be `persistence_id` to match with the journal and snapshot. I think we will use PersistenceId in user api to make it as close to EventSourcedBehavior as possible

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,157 @@
...
+           |  value BYTES(MAX),
```
patriknw, line 29: Possibly we should also include a sequence_nr that is incremented for each update. For eventsByTag I think that is used for deduplication when the the write_time is the same for two events. See SpannerOffset.seen. We will probably need the same if we implement projections for this.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
patriknw: Writer_uuid is mostly for detecting multiple accidental writers. It's probably less useful here since it's in-place updates so difficult to track previous writers. Let's skip that until we know how it would be used.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+/**
+ * INTERNAL API
+ *
+ * Class for doing spanner interaction outside of an actor to avoid mistakes
+ * in future callbacks
+ */
+@InternalApi
```
patriknw, line 55: InternalStableApi since I guess you intend to use this directly

patriknw (commented):
great to se this start

journal/src/main/resources/reference.conf
raboof: possibly. I arrived at 'objects' since while the 'higher-level' features are called 'durable actors'/'value entities', the lower-level storage is just an 'object store'-style API.

Maybe something like 'keyvalue' could also work?

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
raboof: Will do, though we don't actually need it to be *stable*: akka-persistence-spanner is a direct dependency of proxy-spanner, so we are allowed to make breaking changes here (as long as we make sure we can update proxy-spanner accordingly when we upgrade there). Better to be too strict than too loose, though :)

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: Makes sense.

In practice persistenceId's are often composed of an 'identifier per entity type' and a 'entityId to identify a particular instance', do you have thoughts on whether that would be useful to keep in separate colums rather than concatenating them early on? We haven't quite fleshed out what projections we'd like to have, but I can imagine that would come in handy there.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
chbatey: My memory does not go that far back!

chbatey: Me of the past seems to think so: https://github.com/akka/akka-persistence-spanner/pull/15#issuecomment-616520226

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
octonato: When loading the state in memory, we will lookup by persistence id and it will be the composed one (hint + id). I think that splitting them in two columns doesn't add much value. 

And it's possible that users decide to not use a type hint, in that case, splitting will cause the hint column to be empty (null or empty string). Not an issue in itself but we can run into surprises.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,164 @@
...
+    Value((column, value) match {
+      case ((_, TypeCode.STRING), stringValue: String) => StringValue(stringValue)
+      case ((_, TypeCode.INT64), longValue: Long) => StringValue(longValue.toString)
+      case ((_, TypeCode.BYTES), bytes: ByteString) =>
+        // TODO verify why we're sending bytes as base64-encoded strings
+        // there might be some spanner restriction as the API used to be json/http rather than protobuf,
+        // but good to verify
+        StringValue(bytes.encodeBase64.utf8String)
+      case ((_, TypeCode.TIMESTAMP), CommitTimestamp) =>
+        StringValue("spanner.commit_timestamp()")
```
octonato, line 155: Is this asking spanner to take the commit timestamp and add to this column?
Am I understanding this correct?

octonato (commented):
looking good

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: > When loading the state in memory, we will lookup by persistence id and it will be the composed one (hint + id).

Right, there it's no difference indeed.

> it's possible that users decide to not use a type hint

Whether that is possible of course depends on whether we let them :)

My thinking was that perhaps separating those might be useful for projections (though we haven't quite fleshed out yet what those will look like, https://docs.google.com/document/d/1LPNivYNAvj72smhguMVWxBgXs8k4Y47w9Rcfp0ZQtUY/edit#heading=h.yp57xnapyat6 for reference): if you have an application with 2 different types of entities (e.g. 'customers' and 'carts'), would you have 1 projection with all updates to both customer and carts, or separate projections for customers and for carts?

octonato: I would have them as two projections. 

Hmm, I see your point. 
We need a type identifier like in the tagging, so you can query only one of the entities. 

Good point.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
octonato: That's crazy, but I like it. Wondering what kind of injections people can exploit here. 🤔 
They probably have well defined which values can be passed.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
johanandren: Such a field could also be used for optimistic locking/detecting accidental multiple writers as well.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
johanandren: The `commit_timestamp()` is the only special value as far as I know, it requires up front permission in the schema with `allow_commit_timestamp=true` for a `TIMESTAMP` column.

BTW the timestamp is not guaranteed to be unique but has microsecond granularity so it's pretty likely to be unique, so with some extra trixery we are using it for event ordering on the tag side.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
patriknw: but for projections we would need some slicing (sharding) also, so that the load of processing all carts events can be spread over multiple event handlers (nodes). For that we would use tags, but that can be added later.

or do you think we should not support such slicing for these projections?

That said, I don't mind having a separate column for the entity type. We have seen the need for that sometimes and then parsing it out from the persistenceId. The persistenceId column should still be the fully unique key corresponding to the `PersistenceId` in code.

journal/src/main/resources/reference.conf
patriknw: I buy that, `state` might also go away for user api

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -147,9 +147,8 @@ final class SpannerObjectInteractions(
       case ((_, TypeCode.STRING), stringValue: String) => StringValue(stringValue)
       case ((_, TypeCode.INT64), longValue: Long) => StringValue(longValue.toString)
       case ((_, TypeCode.BYTES), bytes: ByteString) =>
-        // TODO verify why we're sending bytes as base64-encoded strings
-        // there might be some spanner restriction as the API used to be json/http rather than protobuf,
-        // but good to verify
+        // the protobuf struct data type doesn't have a type for binary data, so we base64-encode it
+        // https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto#L62
```
johanandren, line 151: 👍

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -44,8 +44,8 @@ private[spanner] final class QuerySettings(config: Config) {
 /**
  * INTERNAL API
  */
-@InternalApi
-private[spanner] final class SpannerSettings(config: Config) {
+@InternalStableApi
```
johanandren, line 47: Internal stable because the settings are used programmatically in your use case?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
```diff
@@ -49,8 +49,10 @@ private[spanner] object SpannerGrpcClient {
 
 /**
  * A thin wrapper around the gRPC client to expose only what the plugin needs.
+ *
+ * INTERNAL API
  */
-@InternalApi private[spanner] class SpannerGrpcClient(
+@InternalStableApi class SpannerGrpcClient(
```
johanandren, line 55: Same question here, is it really used directly in your use case and not only transitively through the new object journal?

journal/src/test/scala/akka/persistence/spanner/PrintSchema.scala
```diff
@@ -41,6 +44,12 @@ object PrintSchema {
       pw.println(SpannerSnapshotInteractions.Schema.Snapshots.snapshotTable(settings))
       pw.println("//#snapshot-tables")
     }
+
+    withWriter("./target/object-tables.txt") { pw =>
+      pw.println("//#object-tables")
+      pw.println(SpannerObjectInteractions.Schema.Objects.objectTable(settings))
```
johanandren, line 50: I thought it was supposed to be internal to start with though, putting the schema in the docs feels quite official/public?

journal/src/test/scala/akka/persistence/spanner/SpannerObjectStoreSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+import akka.actor.typed.scaladsl.adapter.ClassicActorSystemOps
+import akka.persistence.spanner.internal.SpannerObjectInteractions.{ObjectNotFound, Result}
+import akka.persistence.spanner.internal.{SpannerGrpcClientExtension, SpannerObjectInteractions}
+import akka.util.ByteString
+
+class SpannerObjectStoreSpec extends SpannerSpec("SpannerObjectStoreSpec") {
+  override def withObjectStore: Boolean = true
+
+  val grpcClient = SpannerGrpcClientExtension(classicSystem.toTyped).clientFor("akka.persistence.spanner")
+  val spannerInteractions = new SpannerObjectInteractions(
```
johanandren, line 16: Making `Interactions` the user API is a step away from the original intent with them, where we wanted to hide spanner schema specific from the journal and snapshot impls, I think it would be better to call that a `trait SpannerObjectStore` or something like that, potentially backed by the interactions object, and that you can get instances of from an `SpannerObjectStore(system)` or such rather than having to make the GrpcClient semi-public API.

johanandren (commented):
I think the logic looks good, but it would be nice to limit the public/internal stable API surface rather than mark everything as such.

journal/src/test/scala/akka/persistence/spanner/PrintSchema.scala
raboof: we generate this file but don't include this in the docs yet.

journal/src/test/scala/akka/persistence/spanner/SpannerObjectStoreSpec.scala
raboof: The way I was thinking about this is that indeed the journal/snapshot is 'high level' and the Spanner*Interactions are 'low level', and eventually there'd be a 'high level' API for the object store which would be used by both cloudstate and akka-persistence durable actors/entities.

For now I implemented the cloudstate spanner on the 'low level' API's, but already introducing a `SpannerObjectStore(system)` (that would later evolve into a proper akka-persistence api) sounds fine to me - will do.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClient.scala
raboof: there is no object journal yet ;) - but indeed hid it in `SpannerObjectStore` now so it no longer needs to be `InternalStableApi`.

raboof:
(I think I covered all remaining comments, so ready for another look ;) )

journal/src/test/scala/akka/persistence/spanner/PrintSchema.scala
johanandren: Pretty sure we do: https://github.com/akka/akka-persistence-spanner/blob/master/build.sbt#L79

journal/src/test/scala/akka/persistence/spanner/SpannerObjectStoreSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+
+package akka.persistence.spanner
+
+import akka.persistence.spanner.SpannerObjectStore.{ObjectNotFound, Result}
+import akka.util.ByteString
+
+class SpannerObjectStoreSpec extends SpannerSpec("SpannerObjectStoreSpec") {
+  override def withObjectStore: Boolean = true
+
+  val spannerInteractions = SpannerObjectStore()
```
johanandren, line 13: Much nicer 👍

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,177 @@
...
+          case Objects.Key => wrapValue(Objects.Key, key)
+          case Objects.SerId => wrapValue(Objects.SerId, serId)
+          case Objects.SerManifest => wrapValue(Objects.SerManifest, serManifest)
+          case Objects.SeqNr => wrapValue(Objects.SeqNr, seqNr)
+          case Objects.Value => wrapValue(Objects.Value, value)
+          case Objects.WriteTime => wrapValue(Objects.WriteTime, CommitTimestamp)
+          case other => throw new MatchError(other)
+        }))
+      )
+    )
```
johanandren, line 157: Did you consider my rambling about optimistic locking? I'm thinking with blob-storage where we overwrite the whole thing it could be good.

Since there is a seq nr we could chose not to do upsert but only do inserts for seq nr 1 and only updates when it is not, with the current known seq nr as part of the query side making it fail if there was a concurrent writer updating it to another seq nr than we know.

That still leaves some holes for multiple initial instances with the same persistence id though, but perhaps it is worth it anyway for all the other cases?

journal/src/test/scala/akka/persistence/spanner/PrintSchema.scala
raboof: we definitely dump it, but (unlike the other dumps) don't include it in any pages AFAICS

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: > making it fail if there was a concurrent writer updating it to another seq nr than we know.

Hmm, I added the seqnr so we could at least detect it on the read side. I hadn't updated the write side yet - seemed like a bit of a shame to do 2 roundtrips to the db rather than one - but perhaps the added safety is indeed worth it since indeed we overwrite old values.

journal/src/test/scala/akka/persistence/spanner/PrintSchema.scala
johanandren: Hmm, so it was manually added to the docs then. Ok 👍

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
johanandren: I think only the initial insert would potentially require a second roundtrip, for the other ones you'd include the (previous) seqnr in the query side of the update (but now that I think about it, perhaps that is problematic to index in spanner) so that it doesn't match anything and fails if there was an update the writer does not know about.

raboof: > you'd include the (previous) seqnr in the query side of the update (but now that I think about it, perhaps that is problematic to index in spanner)

Ah, not with the Mutations API but looks like the DML API does allow this :+1:

raboof: Must resist writing a mini ORM :laughing:

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,260 @@
...
+            case other => throw new MatchError(other)
+          }), SqlUpdateTypes.mapValues(Type(_)).toMap)
+        )
+      )(session)
+      .map(response => {
+        val updated = response.resultSets.head.stats.head.rowCount.rowCountExact.head
+        if (updated != 1L)
+          throw new IllegalStateException(
+            s"Update failed: object for key [$key] could not be updated to sequence number [$seqNr]"
+          )
```
johanandren, line 159: So we'd end up here on concurrent writer that had updated the seqnr behind our back. Neat 👍

johanandren (commented):
Lookin' good!

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,260 @@
...
+      entity: String,
+      key: String,
+      serId: Long,
+      serManifest: String,
+      value: ByteString,
+      seqNr: Long,
+      session: SessionPool.PooledSession
+  ) =
+    spannerGrpcClient.client
+      .read(
```
patriknw, line 106: I see you discussed the extra roundtrip. I wonder if it's worth it for the first insert. It's anyway not guaranteed to always catch the  problem since the upsert is performed in flatMap and another writer can insert the row between the read and upsert. Is there no strict insert that would fail if row exists?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerSnapshotInteractions.scala
```diff
@@ -282,6 +282,7 @@ private[spanner] final class SpannerSnapshotInteractions(
               )
             )
           )
+          .map(_ => ())
```
patriknw, line 285: ExecutionContexts.parasitic for the last map

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,260 @@
...
+          Mutation(
+            Mutation.Operation.Delete(
+              Mutation.Delete(
+                settings.objectTable,
+                wrapKey(key)
+              )
+            )
+          )
+        )
+      )(session)
```
patriknw, line 207: what happens when trying to delete a key that doesn't exist? should that be a failed Future or a silent success?

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
```diff
@@ -0,0 +1,58 @@
...
+  def deleteObject(key: String): Future[Unit] = interactions.deleteObject(key)
+}
+
+/**
+ * This API would eventually evolve into the akka-persistence object API
+ *
+ * INTERNAL API
+ */
+object SpannerObjectStore {
+  case class ObjectNotFound(key: String) extends Exception(s"No data found for key [$key]")
```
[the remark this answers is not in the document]
patriknw: and NoStackTrace to make it more lightweight

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
```diff
@@ -0,0 +1,58 @@
...
+  def upsertObject(
+      entity: String,
+      key: String,
+      serId: Long,
+      serManifest: String,
+      value: ByteString,
+      seqNr: Long
+  ): Future[Unit] = interactions.upsertObject(entity, key, serId, serManifest, value, seqNr)
+
+  def getObject(key: String): Future[Result] = interactions.getObject(key)
```
patriknw, line 33: Alternative to ObjectNotFound  exception would be Option[Result]. Missing key isn't an error but I don't have a strong opinion about it.

patriknw (commented):
looking good, reviewed in detail

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,260 @@
...
+)(
+    implicit ec: ExecutionContext,
+    system: ActorSystem
+) {
+  import SpannerObjectInteractions.Schema.Objects
+
+  val SqlUpdate =
+    s"UPDATE ${settings.objectTable} SET value = @value, ser_id = @ser_id, ser_manifest = @ser_manifest, " +
+    s"sequence_nr = @new_sequence_nr, write_time = PENDING_COMMIT_TIMESTAMP() " +
+    s"WHERE key = @key AND sequence_nr = @old_sequence_nr"
```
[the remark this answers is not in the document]
raboof: No - I figured it might not be too bad since `key` is the primary key, but that is just an assumption

journal/src/main/resources/reference.conf
viktorklang: `entities`?

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: Mainly adding an `entity` column for the entity type, but keeping the `key` column fully unique - I don't have a strong preference between `key` or `persistence_id` here either

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: great point, there indeed is, updated.

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
raboof: I like that, updating

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
octonato: Seems we decided to not include the `writer_uuid`, so better to remove this commented out lines.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
octonato: We should not need a second roundtrip on insert as it will fail if the row already exists. 

And yes, I find is a good a idea to have built-in optimistic locking if the price to pay is negligible.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
octonato: This was marked as resolved, but still annotated with `@InternalApi`. An oversight?

octonato (commented):
I did another review round and left a few more comments.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: that was intentional, since in 8d2d11911c0eb14b43f4c7cb60eaacb53137748b we introduced the `SpannerObjectStore` API to be used from Akka Serverless, instead of using `SpannerObjectInteractions` directly.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -0,0 +1,247 @@
...
+      )(session)
+      .map(response => {
+        val updated = response.resultSets.head.stats.head.rowCount.rowCountExact.head
+        if (updated != 1L)
+          throw new IllegalStateException(
+            s"Update failed: object for persistence id [$persistenceId] could not be updated to sequence number [$seqNr]"
+          )
+      })
+
+  // TODO maybe return timestamp?
```
patriknw, line 177: I think that can be useful to include in the Result. Correlation with Projections, where it will be the offset.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: > We should not need a second roundtrip on insert as it will fail if the row already exists.

Correct, done in 1e46817e0a0270d3a7b1ab706da7e1e20ef5a6dc

journal/src/test/scala/akka/persistence/spanner/SpannerObjectStoreSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+        .futureValue
+    }
+    "support deletions" in {
+      val persistenceId = entityType + "to-be-added-and-removed"
+      val value = ByteString("Genuinely Collaborative")
+      spannerInteractions.upsertObject(entityType, persistenceId, serId, serManifest, value, 0L).futureValue
+      spannerInteractions.getObject(persistenceId).futureValue should be(Some(Result(value, serId, serManifest, 0L)))
+      spannerInteractions.deleteObject(persistenceId).futureValue
+      spannerInteractions.getObject(persistenceId).futureValue should be(None)
+    }
```
patriknw, line 91: test for delete of already deleted object

patriknw (approved):
LGTM, we can cut another RC

raboof:
I'll still do the `String`->`PersistenceId` type change for what-used-to-be-`key`

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
raboof: Is it, though? A failure that goes unnoticed may cause more problems than some noise...

raboof:
> I'll still do the `String`->`PersistenceId` type change for what-used-to-be-`key`

(done)

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
patriknw: If deletes are only supposed to be performed by the single-writer entity it should know that the object exists or not. In that case it is useful to detect that it was an invalid delete. Otherwise idempotency could be a reason for silently accepting delete of already deleted.

PR #104, opened 2021-02-11
Update copyright to 2021

raboof:


journal/src/test/scala/akka/persistence/spanner/internal/SpannerSerializerSpec.scala
```diff
@@ -1,6 +1,7 @@
 /*
- * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2021 Lightbend Inc. <http://www.lightbend.com>
```
octonato, line 2: curious why that single file was using https.

And is that what we want? Isn't that better to just use https?

octonato: Checked akka/akka and we are using https. Probably also in other projects.

octonato (approved):
LGTM

Left some minor comments, not blocking.

journal/src/test/scala/akka/persistence/spanner/internal/SpannerSerializerSpec.scala
raboof: we should definitely use https - not sure what happened there :)

raboof: (adding `test:headerCheck` to travis)

PR #107, opened 2021-03-22
Added object changes support

jroper:
This adds changes and currentChanges queries to the object store.

This change doesn't change any existing storage structure or format, except that it does add an index - however, given that `akka-persistence-spanner` doesn't actually create any tables/indexes for you, that's up to consumers of the library, the creation of that index is opt-in. The index does create a hotspot on inserting/updating objects for the same entity type.

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -38,10 +38,42 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
 
   def deleteObject(persistenceId: PersistenceId): Future[Unit] = interactions.deleteObject(persistenceId)
 
+  /**
+   * Get a source of the most recent changes made to objects of the given entity type since the passed in offset.
+   *
+   * Note that this only returns the most recent change to each object, if an object has been updated multiple times
+   * since the offset, only the most recent of those changes will be part of the stream.
+   *
+   * This will return changes that occurred up to when the `Source` returned by this call is materialized. Changes to
+   * objects made since materialization are not guaranteed to be included in the results.
+   *
+   * @param entityType The type of entity to get changes for.
+   * @param offset The offset to get changes since. Must either be [[akka.persistence.query.NoOffset]] to get
+   *               changes since the beginning of time, or an offset that has been previously returned by this query.
+   *               Any other offsets are invalid.
+   * @return A source of change events.
+   */
   @ApiMayChange
   def currentChanges(entityType: String, offset: Offset): Source[Change, NotUsed] =
     interactions.currentChanges(entityType, offset)
 
+  /**
+   * Get a source of the most recent changes made to objects of the given entity type since the passed in offset.
+   *
+   * The returned source will never terminate, it effectively watches for changes to the objects and emits changes as
+   * they happen.
+   *
+   * Not all changes that occur are guaranteed to be emitted, this call only guarantees that eventually, the most
+   * recent change for each object since the offset will be emitted. In particular, multiple updates to a given object
+   * in quick succession are likely to be skipped, with only the last update resulting in a change event from this
+   * source.
+   *
+   * @param entityType The type of entity to get changes for.
+   * @param offset The offset to get changes since. Must either be [[akka.persistence.query.NoOffset]] to get
+   *               changes since the beginning of time, or an offset that has been previously returned by this query.
+   *               Any other offsets are invalid.
+   * @return A source of change events.
+   */
   @ApiMayChange
   def changes(entityType: String, offset: Offset): Source[Change, NotUsed] =
     interactions.changes(entityType, offset)
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -1,14 +1,17 @@
 /*
- * Copyright (C) 2020 Lightbend Inc. <http://www.lightbend.com>
+ * Copyright (C) 2021 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.persistence.spanner
 
+import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.actor.typed.scaladsl.adapter.ClassicActorSystemOps
-import akka.annotation.InternalStableApi
+import akka.annotation.{ApiMayChange, InternalStableApi}
+import akka.persistence.query.Offset
 import akka.persistence.spanner.internal.{SpannerGrpcClientExtension, SpannerObjectInteractions}
 import akka.persistence.typed.PersistenceId
+import akka.stream.scaladsl.Source
 import akka.util.ByteString
 
 import scala.concurrent.Future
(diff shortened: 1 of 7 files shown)


jroper:
Note this doesn't solve the problem of how to produce or represent object deletions in the change query. I think that should be a different PR, implemented by setting a deleted flag, possibly with an additional index that includes that flag (hopefully done in such a way that events that are not deleted are not included in that index).

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -52,6 +65,29 @@ object SpannerObjectInteractions {
       /** For use in queries: */
       val OldSeqNr = "old_sequence_nr" -> TypeCode.INT64
       val NewSeqNr = "new_sequence_nr" -> TypeCode.INT64
+
+      case class Object(
```
patriknw, line 69: can we have a different class name for this (confusing to use same as java.lang.Object)

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
```diff
@@ -34,6 +37,14 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
   def getObject(persistenceId: PersistenceId): Future[Option[Result]] = interactions.getObject(persistenceId)
 
   def deleteObject(persistenceId: PersistenceId): Future[Unit] = interactions.deleteObject(persistenceId)
+
+  @ApiMayChange
+  def currentChanges(entityType: String, offset: Offset): Source[Change, NotUsed] =
+    interactions.currentChanges(entityType, offset)
+
+  @ApiMayChange
+  def changes(entityType: String, offset: Offset): Source[Change, NotUsed] =
```
patriknw, line 46: Add docs of the semantics for the offset and that not all intermediate changes will be emitted.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerUtils.scala
```diff
@@ -22,4 +25,13 @@ private[spanner] object SpannerUtils {
...
   val nullValue: NullValue =
     com.google.protobuf.struct.Value.Kind.NullValue(com.google.protobuf.struct.NullValue.NULL_VALUE)
+
+  val SpannerNoOffset = SpannerUtils.unixTimestampMillisToSpanner(0L)
+
+  def toSpannerOffset(offset: Offset) = offset match {
+    case NoOffset => SpannerOffset(SpannerNoOffset, Map.empty)
+    case so: SpannerOffset => so
+    case _ =>
+      throw new IllegalArgumentException(s"Spanner does not support offset type: " + offset.getClass)
```
patriknw, line 35: string interpolation or not, that's the question

patriknw (approved):
LGTM

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
octonato: I agree, but we are calling it `Object` everywhere in this API. 

```scala
SpannerObjectStore.getObject()
SpannerObjectStore.upsertObject()
SpannerObjectStore.deleteObject()
```

We may better move to `Value` since we have settle to that. 

```scala
SpannerValueStore.getValue()
SpannerValueStore.upsertValue()
SpannerValueStore.deleteValue()
```

octonato: Alternatively, we could called it `Payload`. 

(not so convinced myself)

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -221,6 +257,85 @@ private[spanner] final class SpannerObjectInteractions(
...
+            )
+          )
+        ),
+        paramTypes = Map("entity_type" -> Type(TypeCode.STRING), "write_time" -> Type(TypeCode.TIMESTAMP))
+      )
+      .statefulMapConcat(deserializeAndAddOffset(spannerOffset))
+      .mapMaterializedValue(_ => NotUsed)
+  }
+
+  def deserializeAndAddOffset(
```
octonato, line 287:
```suggestion
  private def deserializeAndAddOffset(
```

octonato (approved):
LGTM

There is this open topic of having a class called `Object`. I made some suggestions, but either way is good to me.

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
```diff
@@ -34,6 +37,14 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
   def getObject(persistenceId: PersistenceId): Future[Option[Result]] = interactions.getObject(persistenceId)
 
   def deleteObject(persistenceId: PersistenceId): Future[Unit] = interactions.deleteObject(persistenceId)
+
+  @ApiMayChange
+  def currentChanges(entityType: String, offset: Offset): Source[Change, NotUsed] =
```
raboof, line 42: I wonder if `Offset` is the right API here: the spanner offset happens to contain a timestamp, which is what we want here.

If we ever support different backends, their Offset may not contain a timestamp, but for getting object store streams we would still want to query them by timestamp, right? So perhaps we should have a different `Offset` type for the object store - or just pass in a timestamp explicitly.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerUtils.scala
jroper: lol

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
jroper: I don't see any reason why a backend can't use one offset type for the journal, and another for the object store. But anyway, timestamp is not the only possibility, it would make sense for a relational database to use a sequence number that is populated from a sequence on update, since timestamps are not guaranteed to be unique. For Cassandra, it would make sense to use a time based uuid, since it natively handles that type.

Also note that the offsets being used here in Spanner are not just timestamps, they are timestamps plus a map of persistence ids to sequence numbers, since commit timestamps are not unique, there can be many values that were written at the same timestamp, and processing can fail part way through, so to track the offset we track both the timestamp, and every persistence id seen at that timestamp so far, so when we restart we load from that timestamp, discard the values of persistence ids we've already seen, and go from there.

So the offset abstraction from journals is perfect because it already gives us this, and allows each backend to choose the type best suited to the way it stores data.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
jroper: How about `ObjectRow`, this class is an object that has been parsed directly from a row. Note this class name is internal, we don't expose it in any way.

journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
jroper: Added.

jroper:
I've updated in response to review comments (I can't merge myself).

journal/src/main/scala/akka/persistence/spanner/internal/SpannerUtils.scala
viktorklang: Also, `offset.getClass.getName` :)

PR #129, opened 2021-05-10
fix: exlude akka from scala-steward

raboof:
it is only updated as needed


johanandren (approved):
LGTM, failure is jabba outage, created #138 to fix that

raboof:
(I think the failure is #131)

PR #134, opened 2021-05-27
Publish to sonatype via github actions

raboof:


.travis.yml
```diff
@@ -67,10 +58,6 @@ stages:
   - name: whitesource
     if: repo = akka/akka-persistence-spanner AND ( ( branch = master AND type = push ) OR tag =~ ^v )
```
octonato, line 60: Why not move the whitesource part? 

Or are you planning to do it in small steps?

octonato: oh, I see. There are much more jobs to move.

raboof: no _real_ reason - whitesource is ran on every push to master, I only moved the publishing things for now.

johanandren (approved):

PR #137, opened 2021-06-23
Allow messages to actually be the max size for spanner

johanandren:
Underlying Akka gRPC Netty client limits to 4mb by default, with this PR we can reach 10mb which is the spanner limit

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClientExtension.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClientExtension.scala
@@ -5,12 +5,12 @@
 package akka.persistence.spanner.internal
 
 import java.util.concurrent.ConcurrentHashMap
-
 import akka.Done
 import akka.actor.CoordinatedShutdown
 import akka.actor.typed.scaladsl.adapter._
 import akka.actor.typed.{ActorSystem, Extension, ExtensionId}
 import akka.annotation.InternalApi
+import akka.annotation.InternalStableApi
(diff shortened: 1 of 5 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SpannerGrpcClientExtension.scala
```diff
@@ -42,22 +48,26 @@ private[spanner] class SpannerGrpcClientExtension(system: ActorSystem[_]) extend
...
+                                      GoogleCredentials.getApplicationDefault
+                                        .createScoped("https://www.googleapis.com/auth/spanner.data")
+                                    )
+                                  )
+                              } else {
+                                GrpcClientSettings.fromConfig(settings.grpcClient)
+                              }).withChannelBuilderOverrides(
+          channelBuilder =>
+            // let that max be up to spanner
+            channelBuilder.maxInboundMessageSize(Integer.MAX_VALUE)
```
patriknw, line 68: no drawbacks, such as allocating fixed sized buffers?

johanandren: From what I can see it is only used for the size check (in https://github.com/grpc/grpc-java/blob/d4fa0ecc07495097453b0a2848765f076b9e714c/core/src/main/java/io/grpc/internal/MessageDeframer.java#L387 )

patriknw (approved):
LGTM

pvlugter (approved):

PR #138, opened 2021-06-23
Use JDK 8 from debian package

johanandren:
Avoiding the annoying jabba outages


raboof (approved):
fine by me (though did we see them a lot since #131 ?)

raboof:
I guess this means we now also release with jdk11 rather than jdk8?

johanandren:
Ah, it should perhaps be jdk8 rather, good catch

johanandren:
I don't think there is a huge benefit to testing with jdk11, we can just do all with jdk8

PR #145, opened 2021-07-29
Durable state store.

RayRoestenburg:
Durable state store.
- [x] tests for `SpannerDurableStateStore` plugin.
- [x] docs.

### DIFF
--- a/docs/src/main/paradox/durable-state-store.md
+++ b/docs/src/main/paradox/durable-state-store.md
@@ -1,0 +1,65 @@
+# DurableStateStore plugin
+
+## Schema
+
+The following tables need to be created in the configured database:
+
+@@snip [journal-tables](/target/object-tables.txt) { #journal-tables } 
+
+## Configuration
+
+To activate the durable state store plugin, add the following line to your Akka `application.conf`:
+
+```
+akka.persistence.state.plugin = "akka.persistence.spanner.durable-state-store"
+```
+
+Shared configuration is located under `akka.persistence.spanner.durable-state-store`.
+
+Configuration just for the durable-state-store is under `akka.persistence.spanner`. You will need to set at least:
+
+```
+akka.persistence.spanner {
+    project = <spanner project>
+    instance = <spanner instance>
+    database = <spanner database>
+}
+```
+
+Authentication details will be picked up from the environment using the standard 
+mechanism described [here](https://cloud.google.com/docs/authentication/getting-started).
+   
+### Reference configuration 
+
+The following can be overridden in your `application.conf`
+
+@@snip [reference.conf](/journal/src/main/resources/reference.conf)
+
+### Testing with the Spanner emulator
+
+The [Spanner Emulator](https://cloud.google.com/spanner/docs/emulator). 
+Set the following to connect to it:
+
+```
+akka.persistence.state.plugin = "akka.persistence.spanner.durable-state-store"
+  akka.persistence.spanner {
+    session-pool {
+      # emulator only supports a single transaction at a time
+      max-size = 1
+    }
+    
+    use-auth = false
+ }
+akka.grpc.client.spanner-client {
+  host = localhost
+  port = 9010
+  use-tls = false
+}
+```
+
+You will need to have created the instance and database.
+
+### Deletes
+
+The journal supports deletes through hard deletes, which means the journal entries are actually deleted from the database. 
+There is no materialized view with a copy of the event so if an event that is tagged is deleted it will no longer show up in events by tag queries.
--- a/docs/src/main/paradox/durable-state-store.md
+++ b/docs/src/main/paradox/durable-state-store.md
@@ -4,7 +4,7 @@
 
 The following tables need to be created in the configured database:
 
-@@snip [journal-tables](/target/object-tables.txt) { #journal-tables } 
+@@snip [object-tables](/target/object-tables.txt) { #object-tables } 
 
 ## Configuration
 
@@ -61,5 +61,5 @@ You will need to have created the instance and database.
 
 ### Deletes
 
-The journal supports deletes through hard deletes, which means the journal entries are actually deleted from the database. 
-There is no materialized view with a copy of the event so if an event that is tagged is deleted it will no longer show up in events by tag queries.
+The durable state store supports deletes through hard deletes, which means the durable state store entries are actually deleted from the database. 
+There is no materialized view with a copy of the event so if an event that is tagged is deleted it will no longer show up in changes by tag queries.
--- a/docs/src/main/paradox/durable-state-store.md
+++ b/docs/src/main/paradox/durable-state-store.md
@@ -62,4 +62,4 @@ You will need to have created the instance and database.
 ### Deletes
 
 The durable state store supports deletes through hard deletes, which means the durable state store entries are actually deleted from the database. 
-There is no materialized view with a copy of the event so if an event that is tagged is deleted it will no longer show up in changes by tag queries.
+There is no materialized view with a copy of the state so if a state that is tagged is deleted it will no longer show up in changes by tag queries.
--- a/journal/src/main/resources/reference.conf
+++ b/journal/src/main/resources/reference.conf
@@ -32,6 +32,12 @@ akka.persistence.spanner {
     table = "snapshots"
   }
 
+  #durable-state
+  durable-state-store {
+    class = "akka.persistence.spanner.state.SpannerDurableStateStoreProvider"
+    table = "objects"
+  }
+
   object {
     table = "objects"
   }
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -80,6 +80,46 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
   @ApiMayChange
   def changes(entityType: String, offset: Offset): Source[Change, NotUsed] =
     interactions.changes(entityType, offset)
+
+  /**
+   * Get a source of the most recent changes made to objects of the given tag since the passed in offset.
+   *
+   * Note that this only returns the most recent change to each object, if an object has been updated multiple times
+   * since the offset, only the most recent of those changes will be part of the stream.
+   *
+   * This will return changes that occurred up to when the `Source` returned by this call is materialized. Changes to
+   * objects made since materialization are not guaranteed to be included in the results.
+   *
+   * @param tag The tag to get changes for.
+   * @param offset The offset to get changes since. Must either be [[akka.persistence.query.NoOffset]] to get
+   *               changes since the beginning of time, or an offset that has been previously returned by this query.
+   *               Any other offsets are invalid.
+   * @return A source of change events.
+   */
+  @ApiMayChange
+  def currentChangesByTag(tag: String, offset: Offset): Source[Change, NotUsed] =
+    interactions.currentChanges(tag, offset)
+
+  /**
+   * Get a source of the most recent changes made to objects of the given tag since the passed in offset.
+   *
+   * The returned source will never terminate, it effectively watches for changes to the objects and emits changes as
+   * they happen.
+   *
+   * Not all changes that occur are guaranteed to be emitted, this call only guarantees that eventually, the most
+   * recent change for each object since the offset will be emitted. In particular, multiple updates to a given object
+   * in quick succession are likely to be skipped, with only the last update resulting in a change event from this
+   * source.
+   *
+   * @param tag The tag to get changes for.
+   * @param offset The offset to get changes since. Must either be [[akka.persistence.query.NoOffset]] to get
+   *               changes since the beginning of time, or an offset that has been previously returned by this query.
+   *               Any other offsets are invalid.
+   * @return A source of change events.
+   */
+  @ApiMayChange
+  def changesByTag(tag: String, offset: Offset): Source[Change, NotUsed] =
+    interactions.changes(tag, offset)
 }
 
 /**
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -109,7 +109,7 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
    */
   @ApiMayChange
   def currentChangesByTag(tag: String, offset: Offset): Source[Change, NotUsed] =
-    interactions.currentChanges(tag, offset)
+    interactions.currentChangesByTag(tag, offset)
 
   /**
    * Get a source of the most recent changes made to objects of the given tag since the passed in offset.
@@ -130,7 +130,7 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
    */
   @ApiMayChange
   def changesByTag(tag: String, offset: Offset): Source[Change, NotUsed] =
-    interactions.changes(tag, offset)
+    interactions.changesByTag(tag, offset)
 }
 
 /**
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -37,6 +37,17 @@ class SpannerObjectStore(interactions: SpannerObjectInteractions) {
       seqNr: Long
   ): Future[Unit] = interactions.upsertObject(entityType, persistenceId, serId, serManifest, value, seqNr)
 
+  @ApiMayChange
+  def upsertObject(
+      entityType: String,
+      persistenceId: PersistenceId,
+      serId: Long,
+      serManifest: String,
+      value: ByteString,
+      seqNr: Long,
+      tag: String
+  ): Future[Unit] = interactions.upsertObject(entityType, persistenceId, serId, serManifest, value, seqNr, tag)
+
   def getObject(persistenceId: PersistenceId): Future[Option[Result]] = interactions.getObject(persistenceId)
 
   def deleteObject(persistenceId: PersistenceId): Future[Unit] = interactions.deleteObject(persistenceId)
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
@@ -96,7 +96,8 @@ object SpannerObjectStore {
       serId: Long,
       serManifest: String,
       seqNr: Long,
-      offset: Offset
+      offset: Offset,
+      timestamp: Long
   )
 
   def apply()(implicit system: ActorSystem): SpannerObjectStore = {
--- a/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
@@ -9,6 +9,7 @@ import akka.persistence.spanner.SpannerSettings.SessionPoolSettings
 import akka.util.JavaDurationConverters._
 import com.typesafe.config.Config
 
+import scala.util.Try
 import scala.concurrent.duration.FiniteDuration
 
 /**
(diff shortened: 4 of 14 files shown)


journal/src/main/scala/akka/persistence/spanner/SpannerObjectStore.scala
```diff
@@ -96,7 +96,8 @@ object SpannerObjectStore {
       serId: Long,
       serManifest: String,
       seqNr: Long,
-      offset: Offset
+      offset: Offset,
+      timestamp: Long
```
RayRoestenburg, line 100: Added timestamp, so SpannerDurableStateStore can use it, probably a breaking change for existing users.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -302,7 +302,8 @@ private[spanner] final class SpannerObjectInteractions(
           serId = obj.serId,
           serManifest = obj.serManifest,
           seqNr = obj.sequenceNr,
-          offset = offset
+          offset = offset,
+          timestamp = SpannerUtils.spannerTimestampToUnixMillis(obj.writeTime)
```
RayRoestenburg, line 306: Added timestamp.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  private def deserialize(bytes: Array[Byte], serId: Int, serManifest: String) =
+    serialization
+      .deserialize(
+        bytes,
+        serId,
+        serManifest
+      )
+      .map(_.asInstanceOf[A])
+}
```
debasishg, line 112: should we think of a serialization class where this can be refactored into ? We have  a similar `AkkaSerialization` class in JDBC plugin ..

RayRoestenburg: Different projects, I would prefer we don't do that.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+
+  def upsertObject(persistenceId: String, revision: Long, value: A, tag: String): Future[Done] =
+    for {
+      (bytes: ByteString, serId: Int, serManifest: String) <- Future.fromTry(
+        serialize(value)
+      )
+      _ <- spannerObjectStore.upsertObject(
+        tag, // tag is used as entityType in spanner store
```
RayRoestenburg, line 66: I assumed this, please check. (tag == entityType)

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+        bytes,
+        revision
+      )
+    } yield Done
+
+  def deleteObject(persistenceId: String): Future[Done] =
+    spannerObjectStore.deleteObject(PersistenceId.ofUniqueId(persistenceId)).map(_ => Done)
+
+  def currentChanges(tag: String, offset: Offset): Source[DurableStateChange[A], NotUsed] =
+    // tag == entityType in spanner store
```
RayRoestenburg, line 79: assumption.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+    spannerObjectStore.currentChanges(tag, offset).map(change => toDurableStateChange(change))
+
+  def changes(tag: String, offset: Offset): Source[DurableStateChange[A], NotUsed] =
+    // tag == entityType in spanner store
```
RayRoestenburg, line 83: assumption.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
RayRoestenburg: This assumption allows me to completely reuse SpannerObjectStore, if this is not wanted, I guess we will have to add another table definition specifically with a "tag" column in it, or possibly, add a tag column? The issue with that would be that you can't always extract entity out of persistenceId string.. help please.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
debasishg: I am not talking about reusing. Maybe we can factor out the serialization bits into a similar class here named `AkkaSerialization`

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+    with DurableStateStoreQuery[A] {
+  def getObject(persistenceId: String): Future[GetObjectResult[A]] =
+    spannerObjectStore.getObject(PersistenceId.ofUniqueId(persistenceId)).flatMap { result =>
+      result match {
+        case Some(r) =>
+          for {
+            deserialized <- Future
+              .fromTry(
+                // to Int: Spanner only has INT64
+                deserialize(r.byteString.toArray, r.serId.asInstanceOf[Int], r.serManifest)
```
debasishg, line 48: Is `r.serId` a Long ? Then we can use `r.serId.toInt` instead of the cast.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+  def changes(tag: String, offset: Offset): Source[DurableStateChange[A], NotUsed] =
+    // tag == entityType in spanner store
+    spannerObjectStore.changes(tag, offset).map(change => toDurableStateChange(change))
+
+  private def toDurableStateChange(change: SpannerObjectStore.Change) =
+    new DurableStateChange[A](
+      persistenceId = change.persistenceId,
+      revision = change.seqNr,
+      // to Int: Spanner only has INT64
+      value = deserialize(change.bytes.toArray, change.serId.asInstanceOf[Int], change.serManifest).get, // crash source if corrupt.
```
debasishg, line 91: same as before `r.serId.toInt` instead of the cast ?

journal/src/test/scala/akka/persistence/spanner/SpannerDurableStateStoreSpec.scala
```diff
@@ -0,0 +1,137 @@
...
+import akka.persistence.spanner.SpannerObjectStore.{Change, Result}
+import akka.persistence.state.DurableStateStoreRegistry
+import akka.persistence.state.scaladsl.GetObjectResult
+import akka.persistence.typed.PersistenceId
+import akka.stream.scaladsl.Sink
+import akka.stream.testkit.scaladsl.TestSink
+import akka.util.ByteString
+
+class SpannerDurableStateStoreSpec extends SpannerSpec("SpannerDurableStateStoreSpec") {
+  override def withObjectStore: Boolean = true
```
debasishg, line 20: can't we make this a `val` ?

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
RayRoestenburg: oh ok, got it. Everything is already in two small methods, `serialize` and `deserialize`, should be ok

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
RayRoestenburg: So instead, I'm going to add a `tag` column in the table, and always fill `entityType` with empty string. I'll add a ticket for further improvement. There are likely follow up discussions around `PersistenceId`.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
RayRoestenburg: Of course! argh

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
RayRoestenburg: Don't rub it in please 😅

journal/src/test/scala/akka/persistence/spanner/SpannerDurableStateStoreSpec.scala
RayRoestenburg: yeah probably

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -40,6 +40,7 @@ object SpannerObjectInteractions {
            |  ser_manifest STRING(MAX) NOT NULL,
            |  sequence_nr INT64 NOT NULL,
            |  write_time TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
+           |  tag STRING(MAX) NOT NULL
```
RayRoestenburg, line 43: Added tag column so durable state store works, and SpannerObjectStore continues to work as is, though it does mean a table schema change. Expecting more improvements in this area (tagging, dynamic projections, etc).

ignasi35: Maybe it should be nullable? 🤔

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
```diff
@@ -131,15 +144,26 @@ private[spanner] final class SpannerObjectInteractions(
...
 
+  def upsertObject(
+      entity: String,
+      persistenceId: PersistenceId,
+      serId: Long,
+      serManifest: String,
+      value: ByteString,
+      seqNr: Long
+  ): Future[Unit] =
+    upsertObject(entity, persistenceId, serId, serManifest, value, seqNr, "")
```
ignasi35, line 165: Oh, the empty tag is not `Null` but `""`.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -63,7 +63,7 @@ class SpannerDurableStateStore[A](
         serialize(value)
       )
       _ <- spannerObjectStore.upsertObject(
-        tag, // tag is used as entityType in spanner store
+        "", // cannot extract entityType from persistenceId.
```
ignasi35, line 66: Should this `upsertObject()` invocation set the tag? (use the new overload)

ignasi35 (changes requested):
upsertObject must pass the tag.

ignasi35 (approved):
I was reviewing commit by commit and noticed you already added the fix.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
RayRoestenburg: I fixed it, yes, this sets the tag.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+              .fromTry(
+                // to Int: Spanner only has INT64
+                deserialize(r.byteString.toArray, r.serId.asInstanceOf[Int], r.serManifest)
+              )
+          } yield {
+            GetObjectResult(
+              Some(deserialized),
+              r.seqNr
+            )
+          }
```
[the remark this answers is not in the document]
RayRoestenburg: yes in this case it's not the best use of it ;-)

ignasi35:
In other persistence plugins, when changing the schema we either document or provide the migration tools for users that already deployed the old schema.

We should have the `update` DML or docs alerting about the new column.

🤔 Hmm,  I guess it depends on when was the `state` table introduced in the schema.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
```diff
@@ -0,0 +1,112 @@
...
+import akka.persistence.query.scaladsl.DurableStateStoreQuery
+import akka.persistence.spanner.SpannerObjectStore
+import akka.persistence.typed.PersistenceId
+import akka.serialization.Serialization
+import akka.serialization.Serializers
+import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.Source
+import akka.stream.Materializer
+import akka.stream.SystemMaterializer
+import akka.util.Timeout
```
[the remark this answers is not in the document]
RayRoestenburg: why no warning in compile, are you sure?

journal/src/main/resources/reference.conf
```diff
@@ -32,6 +32,12 @@ akka.persistence.spanner {
...
 
+  #durable-state
+  durable-state-store {
+    class = "akka.persistence.spanner.state.SpannerDurableStateStoreProvider"
+    table = "objects"
+  }
+
   object {
     table = "objects"
   }
```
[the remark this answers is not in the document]
RayRoestenburg: same table.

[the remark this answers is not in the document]
RayRoestenburg: I guess I can remove the `table=objects` in `durable-state-store`

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
RayRoestenburg: just adding in empty, the implementation of DurableStateStore expects it to exist.

journal/src/main/resources/reference.conf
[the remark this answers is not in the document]
RayRoestenburg: Its actually a good point.. I have now added it tries to get the table first from durable-state-store, fallback to store.

raboof:
> In other persistence plugins, when changing the schema we either document or provide the migration tools for users that already deployed the old schema.
> 
> We should have the `update` DML or docs alerting about the new column.
> 
> thinking Hmm, I guess it depends on when was the `state` table introduced in the schema.

In this case we've never released a non-RC version with this table, so changing without docs. It might be nice to document it for akkaserverless' sake, though :)

journal/src/main/resources/reference.conf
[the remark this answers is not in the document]
raboof: Perhaps we should remove `object` alltogether and put the options under `durable-state-store`? I don't see a reason to keep both.

debasishg (approved):
LGTM

docs/src/main/paradox/durable-state-store.md
```diff
@@ -0,0 +1,65 @@
...
+  use-tls = false
+}
+```
+
+You will need to have created the instance and database.
+
+### Deletes
+
+The durable state store supports deletes through hard deletes, which means the durable state store entries are actually deleted from the database. 
+There is no materialized view with a copy of the event so if an event that is tagged is deleted it will no longer show up in changes by tag queries.
```
ignasi35, line 65:
```suggestion
There is no materialized view with a copy of the state so if a state that is tagged is deleted it will no longer show up in changes by tag queries.
```

Hmmm, this sounds dangerous. If I'm reading correctly, a projection may loose the last few changes of an entity:

```
...
entity123-rev6
entity123-rev7
entity123-rev8
entity123-rev9
(deleted)
```

If the projection lags behind and only consumes up to `entity123-rev7`, hard-deleting the entity at rev10 means the projection will not only miss rev8 and rev9, but also it'll miss the fact that the entity was deleted.

Is that alright? It looks like projections will become out of sync.

ignasi35: This problem is probably not exclusive to the spanner plugin, though. How do deletes work in the JDBC plugin and how does the JDBC plugin prevent this problem? cc @debasishg

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -68,7 +69,8 @@ final class SpannerSettings(config: Config) {
   val snapshotsTable = config.getString("snapshot.table")
 
   // Object store to (eventually) back durable actors
-  val objectTable = config.getString("object.table")
+  val objectTable =
```
ignasi35, line 72: Is there a reason to not drop `object.table` yet?

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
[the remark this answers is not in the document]
ignasi35: Maybe we haven't set up the disclipline plugin or the compiler settings. The IDE dropped a good chuck...

journal/src/main/resources/reference.conf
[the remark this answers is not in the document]
RayRoestenburg: sure can do that.

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
raboof: Instead of having both the `entityType` and the `tag`, perhaps we should just *replace* `entityType` with `tag`?

AFAIK the main reason for having `entityType` in Akka Serverless is so that we can then query for that entity in views - but that could work exactly the same by tagging all entities of a certain entity type with the same tag.

raboof (commented):
In the end is it still useful to have both `SpannerObjectStore` and `scaladsl/SpannerDurableStateStore`? Eventually we'd want to port akkaserverless-framework to use the more public Durable State API's anyway. Perhaps keeping it for now might make porting easier, though

RayRoestenburg:
@raboof yes I'm leaving the existing one for now for compatibility, and there are still some issues / differences.

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
RayRoestenburg: The existing API uses that (expects that) so it is likely more compatible like this now. I think it should be removed very soon, there will be a follow up issue.

raboof:
> there are still some issues / differences.

What are they? Should we resolve these before merging?

journal/src/main/scala/akka/persistence/spanner/state/scaladsl/SpannerDurableStateStore.scala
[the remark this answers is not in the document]
RayRoestenburg: ah discipline, is that a flag? maybe I forgot to run sbt with it if this is not automatically enabled...

journal/src/main/resources/reference.conf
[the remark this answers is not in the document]
RayRoestenburg: Let's do that in follow up where we don't have the other implementation, since this config is probably used by existing apps.

RayRoestenburg:
@raboof there is a wider conversation probably to be had about `PersistenceId`, entityType / Hint, being able to extract this or not, supporting only persistence typed or not, but this conversation is probably best done when some are back from holiday. I'll add a placeholder for now.

journal/src/main/scala/akka/persistence/spanner/internal/SpannerObjectInteractions.scala
patriknw: Ouch, in Akka Serverless there is no slices/tags yet so this is in the `entity_type` column. That was rather stupid of us to not add the tag column up front. This will be some migration work now.

patriknw (commented):
We should try this out before releasing next version. I suspect several thing in the implementation can be simplified. SpannerDurableStateStore and SpannerObjectStore can probably be collapsed.

PR #149, opened 2021-08-17
Simplify SpannerDurableStateStore

patriknw:
* Remove the old SpannerObjectStore
* Remove old config
* Make the tag column name configurable so that it can be
  set to entity_type for compatibility with previous implementation

This is a follow up on https://github.com/akka/akka-persistence-spanner/pull/145


raboof (approved):
Great to be able to clean up the 'temporary' API! Good that you were able to carry over some comments and tests.

PR #155, opened 2021-09-13
Slice instead of tags

patriknw:
I had a go at "dynamic slices" that we have talked so much about...

* to allow for more dynamic slicing of events in projections
* events are written with their slice id, 0 until 128 (hash modulo of persistenceId)
* events are queried with slice range
  * for example, if starting with 4 projection instances it would be ranges
    0 to 31, 32 to 63, 64 to 95, 96 to 128
  * when scaling out is needed the offset for these ranges can be copied
    and starting more projection instances with smaller ranges, such as
    0 to 15, 16 to 31, ...

### DIFF
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SliceUtils.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SliceUtils.scala
@@ -1,0 +1,34 @@
+/**
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+package akka.persistence.spanner.internal
+
+import akka.annotation.InternalApi
+import akka.persistence.typed.PersistenceId
+import scala.collection.immutable
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[spanner] object SliceUtils {
+  def sliceForPersistenceId(persistenceId: String, maxNumberOfSlices: Int): Int =
+    math.abs(persistenceId.hashCode % maxNumberOfSlices)
+
+  def extractEntityTypeHintFromPersistenceId(persistenceId: String): String = {
+    val i = persistenceId.indexOf(PersistenceId.DefaultSeparator) // TODO configurable separator
+    if (i == -1) ""
+    else persistenceId.substring(0, i)
+  }
+
+  def sliceRanges(numberOfRanges: Int, maxNumberOfSlices: Int): immutable.Seq[Range] = {
+    val rangeSize = maxNumberOfSlices / numberOfRanges
+    require(
+      numberOfRanges * rangeSize == maxNumberOfSlices,
+      s"numberOfRanges [$numberOfRanges] must be a whole number divisor of maxNumberOfSlices [$maxNumberOfSlices]."
+    )
+    (0 until numberOfRanges).map { i =>
+      (i * rangeSize until i * rangeSize + rangeSize)
+    }.toVector
+  }
+}
--- a/journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
+++ b/journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
@@ -20,9 +20,10 @@ import akka.util.ConstantFun
 import com.google.protobuf.struct.Value.Kind.StringValue
 import com.google.protobuf.struct.{ListValue, Struct, Value}
 import com.google.spanner.v1.{Mutation, Type, TypeCode}
-
 import scala.collection.immutable
 import scala.concurrent.{ExecutionContext, Future}
+
+import akka.persistence.typed.PersistenceId
 
 /**
  * INTERNAL API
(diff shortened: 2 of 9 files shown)


journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
```diff
@@ -48,18 +49,31 @@ private[spanner] object SpannerJournalInteractions {
...
            |) PRIMARY KEY (persistence_id, sequence_nr)""".stripMargin
 
+      // FIXME shall include slice, entity_type_hint as first part of primary key, to collocate with slice index?
+
+      // FIXME verify performance of this index, it can't INTERLEAVE IN journal
+      // because then it would have to include persistence_id, sequence_nr and
+      // those are not part of the query
+      def sliceIndex(settings: SpannerSettings): String =
+        s"CREATE INDEX ${settings.journalTable}_slice " +
+        s"ON ${settings.journalTable}(slice, entity_type_hint, write_time)"
```
patriknw, line 71: I have tried this with a real Spanner and it's looking pretty good. I updated and used the `example` in this project. Populated the database with > 1 million events by running it several times with:
```
"-Dakka.cluster.roles.0=write -Dakka.cluster.roles.1=load -Dakka.cluster.roles.2=read -Dakka.cluster.roles.3=report -Dspanner.example.processors=4 -Dspanner.example.persistence-ids=10000 -Dspanner.example.load-tick-duration=100ms -Dspanner.example.events-per-tick=10"
```

I used the query analyzer. A query at the end of the tail looks like this:
```
SELECT *
  FROM test_journal@{FORCE_INDEX=test_journal_slice}
  WHERE slice BETWEEN 0 AND 31
  AND entity_type_hint = 'configurable'
  AND write_time >= '2021-09-16T12:18:22.103871Z'
  ORDER BY write_time, persistence_id, sequence_nr
  LIMIT 1000;
```

tail

A query that is far behind (240k rows behind the tail):
```
SELECT *
  FROM test_journal@{FORCE_INDEX=test_journal_slice}
  WHERE slice BETWEEN 0 AND 31
  AND entity_type_hint = 'configurable'
  AND write_time >= '2021-09-16T09:48:48.103871Z'
  ORDER BY write_time, persistence_id, sequence_nr
  LIMIT 1000;
```

far-behind

It is important to use the force index, otherwise it will do a full table scan like this:

tail-without-index

journal/src/main/scala/akka/persistence/spanner/internal/SliceUtils.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import akka.annotation.InternalApi
+import akka.persistence.typed.PersistenceId
+import scala.collection.immutable
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[spanner] object SliceUtils {
```
patriknw, line 14: this should be public, move to SpannerReadJournal

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
johanandren: Nice, dominated by the distributed union, so I guess there is no way to get it _much_ faster than that.

patriknw: It is possible to store everything needed in the index, which means that it would only have to read from the index. The index becoming a materialized view. However, means duplicate storage cost so I don't think we want to go there.

johanandren (commented):
Nice!

journal/src/main/scala/akka/persistence/spanner/internal/SpannerJournalInteractions.scala
patriknw: That was my thinking. Secondary indexes are stored in separate index tables, treated as actual tables. With the slice first it's more obvious that it should slit it and not have a hot spot when writing to the same entity type.

I would have preferred if there was a way to collocate the index with the journal table. Interleaved index doesn't work because then the persistenceId and seqNr must be the first two columns in the index.

I was thinking it might by chance collocate if we changed the primary key of the journal to include slice, entity_type_hint as first columns (same as index), but I guess splits are based on other things than just number of rows. It would also be more difficult for us to change the primary key (alter primary key is not possible).

octonato: I think we should experiment with bitmask and with a mod function (if supported in Spanner) as well to see how it performs compared with `between`. 

The experiments we did with postgres were looking better if we compare the execution time. 

For a lag of 1 million events, postgres was answering the query in about 180ms, compared to a lag of 250k and exec time of 300ms in Spanner. 

And for a lag of 2000 events, postgres reporting 0,7ms. 

I'm not suggesting we should use Postgres, my point is that I get the feeling that a bitmask or mod function will be more efficient than a between clause. 

The bitmask will offer the same properties when it comes to resizing the number of projections.

PR #172, opened 2021-11-10
paged currentPersistenceIds query

### LINKED ISSUE #170: implement durable state currentPersistenceIds query
Introduced in https://github.com/akka/akka/pull/30811

raboof:
Drafs since we should probably wait for an Akka release?

I see r2dbc has added the interface to the project itself in
https://github.com/akka/akka-persistence-r2dbc/tree/main/core/src/main/scala/akka/persistence/query/scaladsl
but I'm not sure we want to do that here as well?

Fixes #170


patriknw (approved):
LGTM

patriknw:
> Drafs since we should probably wait for an Akka release?

I think we can use the same Akka snapshot version as we are using in Akka Serverless. Maybe not publish a RC for that but use a snapshot of this plugin in Akka Serverless also.

I will start moving pieces from r2dbc plugin to Akka so that we can have a snapshot or release that contains everything.

(btw, RC was a premature decision but we have to live with that)

patriknw (approved):
LGTM

PR #185, opened 2021-12-20
Increase limit for session-pool.max-size, and increase default

patriknw:


journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
```diff
@@ -19,9 +19,8 @@ import scala.concurrent.duration.FiniteDuration
 object SpannerSettings {
   final class SessionPoolSettings(config: Config) {
     val maxSize = config.getInt("max-size")
-    // Spanner only supports 100 sessions per gRPC channel. We'd need multiple channels to support
-    // more
-    require(maxSize <= 100, "session-pool.max-size must be <= 100")
+    // We'd probably need multiple channels to support more
+    require(maxSize <= 500, "session-pool.max-size must be <= 500")
```
johanandren, line 23: I wonder if that changed since we wrote the plugin, sounds weird that we would have just made a number up.

johanandren (approved):

journal/src/main/scala/akka/persistence/spanner/SpannerSettings.scala
patriknw: I think that could have changed.

PR #201, opened 2022-02-15
Move to GH actions

ennru:


.github/workflows/check-build-test.yml
```diff
@@ -0,0 +1,95 @@
...
+      JAVA_OPTS: -Xms2G -Xmx2G -Xss2M -XX:ReservedCodeCacheSize=256M -Dfile.encoding=UTF-8
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with: # https://github.com/olafurpg/setup-scala#faster-checkout-of-big-repos
+          fetch-depth: 100
+
+      - name: Fetch tags
+        run: git fetch --depth=100 origin +refs/tags/*:refs/tags/*
```
patriknw, line 83: does this merge with main? in another repo I have:
```
      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 0

      - name: Checkout GitHub merge
        if: github.event.pull_request
        run: |-
          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
          git checkout scratch
```

.github/workflows/check-build-test.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: "Create all API docs and create site with Paradox"
+        run: sbt "root/doc; docs/makeSite"
+
+      - name: Run Link Validator
+        run: cs launch net.runne::site-link-validator:0.2.2 -- project/link-validator.conf
+
+  test:
+    name: Test with the Spanner emulator
+    runs-on: ubuntu-20.04
```
patriknw, line 71: I think all of them should have
```
if: github.repository == 'akka/akka-persistence-spanner'
```

.github/workflows/check-build-test.yml
ennru: Right, I changed the checkout for tests to use the merge.

.github/workflows/check-build-test.yml
ennru: All but releasing should be OK to run on forks, as well.

patriknw: > All but releasing should be OK to run on forks, as well.

Fine, but probably surprising for contributors and not very environmental friendly.

patriknw (approved):
LGTM

